PR #5, opened 2017-07-10
tensorflow r1.2 initial commit

ppeeou:
Thank you for listening to my request :)


lantiga:
Thank you for your contribution!

If things are working we can merge to master. Also, if you have interesting examples it would be nice to have them in.

PR #8, opened 2018-09-07
replace uname with "shell uname"

gkorland:


lantiga:
Merged, thanks @gkorland

PR #19, opened 2018-10-04
fix #17 number of arguments for FT.TENSOR + VALUES

### LINKED ISSUE #17: Error on more than on dimension  input
```
127.0.0.1:6379> TF.TENSOR gg3 FLOAT 3 1 2 2 VALUES 2 2 2 2
(error) ERR wrong number of arguments for 'TF.TENSOR' command
```

It seems the bug is in
```
398:  if (hasdata && datafmt == REDISTF_DATA_VALUES && argc != len + 6) {
```

Should be:
```
  if (hasdata && datafmt == REDISTF_DATA_VALUES && argc != len + 5 + ndims) {

```

gkorland:


lantiga:
Whoops, my bad. Thanks for the fix

PR #21, opened 2018-10-23
Initial commands documentation

gkorland:


lantiga:
Thank you @gkorland 

Moving forward I'd like to redesign the API to make it more verb-oriented, like DL.TSET, DL.TGET, DL.GSET, DL.GRUN, etc. What do you think?

PR #23, opened 2018-11-28
Small refactoring stage 1

MeirShpilraien:
Added Tensor.c, Tensor.h, Graph.c and Graph.h
Put basic functionalities to the correct components


MeirShpilraien:
@lantiga pls notice that I added a refCount to Tensor struct so we will be able to avoid cloning it in the future (but I did not implemented it yet so its still cloned but I believe we might be able to remove it)

lantiga:
Looks good. Merging, thanks!

PR #40, opened 2019-02-10
Add license badge

gkorland:


lantiga:
Build fails due to `/workspace` not being created in CI, this merge is good but we’ll need to address that asap.

PR #45, opened 2019-02-12
Libtorch backend

lantiga:


src/graph.c
```diff
@@ -33,7 +37,7 @@ int RAI_GraphInit(RedisModuleCtx* ctx) {
       .digest = NULL
   };
 
-  RedisAI_GraphType = RedisModule_CreateDataType(ctx, "DL__GRAPH", 0, &tmGraph);
+  RedisAI_GraphType = RedisModule_CreateDataType(ctx, "AI__GRAPH", 0, &tmGraph);
```
MeirShpilraien: if you change the datatype name it will not be backward compatible, is it ok?

src/graph.c
```diff
@@ -54,6 +61,9 @@ void RAI_GraphFree(RAI_Graph* graph) {
   if (graph->backend == RAI_BACKEND_TENSORFLOW) {
     RAI_GraphFreeTF(graph);
   }
+  else if (graph->backend == RAI_BACKEND_TORCH) {
+    RAI_GraphFreeTorch(graph);
+  }
   else {
     // TODO: err properly
     printf("ERR: Unsupported backend.\n");
```
MeirShpilraien: maybe we should put an assert it cause this really should not happened right?

src/graph.c
```diff
@@ -42,6 +46,9 @@ RAI_Graph *RAI_GraphCreate(RAI_Backend backend, RAI_Device device,
   if (backend == RAI_BACKEND_TENSORFLOW) {
     return RAI_GraphCreateTF(backend, device, graphdef, graphlen);
   }
+  else if (backend == RAI_BACKEND_TORCH) {
+    return RAI_GraphCreateTorch(backend, device, graphdef, graphlen);
+  }
 
   return NULL;
```
MeirShpilraien: could this happened? if not assert?

src/graph.c
lantiga: I don't think we have actual users outside a tiny group. Also, backward compatibility here applies to databases being restored from RDB or AOF right? In that case don't save to those yet.

src/graph.c
lantiga: You mean the unsupported backend, right? I agree

src/graph.c
lantiga: I get the previous comment now: Torch graphs are fully supported. See in `pytorch-minimal.py` and `load_torch_model.sh` in `examples/models` for an example on how to trace+export a graph from PyTorch and run it in Redis. See here for reference: https://pytorch.org/tutorials/advanced/cpp_export.html

src/graph.c
lantiga: (I got your comment now, see next response to comment)

src/graph.c
MeirShpilraien: I actually meant the return NULL if we reach there its a bug right? So it might be better to put an assert there so we will now it happened.

src/tensor.c
```diff
@@ -133,20 +179,28 @@ size_t RAI_TensorGetDataSize(const char* dataTypeStr){
 
 void RAI_TensorFree(RAI_Tensor* t){
   if (--t->refCount <= 0){
-    RedisModule_Free(t->tensor.shape);
-    RedisModule_Free(t->tensor.data);
+    if (t->tensor.manager_ctx && t->tensor.deleter) {
+      t->tensor.deleter(&t->tensor);
+    }
+    else {
+      RedisModule_Free(t->tensor.dl_tensor.shape);
```
MeirShpilraien: I suggest to switch to a macro of alloc and free and when compile for production point macro to use RedisModule_Free and when compile for testing point macro to use std malloc and free, this way it will be easy to run valgrind on it and test for leaks.

MeirShpilraien: Look how we do it in Redisearch

src/graph.c
lantiga: Oh right, sorry. Yes, great point.

src/graph.c
MeirShpilraien: Also I would suggest to create a single graph struct with function pointers to Free, Run , .. and all the other capabilities. This way we can avoid all those if's and it will be easy to add more backends in the future ..

src/tensor.c
lantiga: Leaks is something we should definitely test at this point, I'll do that before merging

lantiga:
Great, thank you @MeirShpilraien!

src/tensor.c
lantiga: This is it right? https://github.com/RedisLabsModules/RediSearch/blob/master/src/rmutil/alloc.h

MeirShpilraien: yes and you should call RMUTil_InitAlloc if REDIS_MODULE_TARGET is not defined so it will switch the pointers

src/graph.c
lantiga: Correct, we should definitely do some refactoring after this round of changes. I'm opening a dedicated issue now.

src/tensor.c
lantiga: Where do you suggest is the best place to call it?

MeirShpilraien: at the init function?

lantiga: Done

src/graph.c
lantiga: Done

lantiga:
@MeirShpilraien ready to merge if you approve. Thank you for the reviews!

PR #52, opened 2019-02-19
Command refactor

lantiga:
This PR brings in a few significant changes:

- Graph is renamed to Model
- Subcommands are gone, commands are now following _objectverb_ (e.g. `TENSORSET`)
- Commands and return values do not include counters (e.g. `ndim dim1 dim2` is now `dim1 dim2`)
- `MODELSET` now requires names in `INPUTS` and `OUTPUTS` for the TF backend
- as a consequence, `MODELRUN` only needs keys

TODO before merging
- [x] ~~thorough testing of the unhappy paths~~ [opened separate issue, see #62]
- [x] evaluate adding placement (it could be optional) in `TENSORSET`
- [x] consider changing the placement tag from `GPU` to `CUDA` (to allow to differentiate devices in the future)


lantiga:
A comment on the open points:
- I considered switching `GPU` to `CUDA`; indeed, dlpack uses `GPU`, as does TF; we'll keep it this way, so no change here
- `TENSORSET` will get an additional placement argument in the future, but the default will be `CPU`, so we can defer adding the placement argument for now

src/redisai.c
```diff
@@ -426,27 +418,67 @@ int RedisAI_Set_Graph_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
+
+  size_t namelen;
+  const char *name, *option;
+
+  option = RedisModule_StringPtrLen(argv[argidx], NULL);
+
+  if (strcasecmp(option, "INPUTS") == 0) {
+    argidx++;
+    for (size_t i=0; i<argc-argidx-1; i++) {
+      name = RedisModule_StringPtrLen(argv[argidx + i], &namelen);
```
mnunberg: It's safe to assume here that the return value (RedisModule_StringPtrLen) is NUL-terminated; no need for the extra alloc/copy here... and the function you're passing this into does not modify the strings.

src/redisai.c
```diff
@@ -426,27 +418,67 @@ int RedisAI_Set_Graph_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
+    }
+    argidx += array_len(inputs);
+  }
+
+  option = RedisModule_StringPtrLen(argv[argidx], NULL);
+
+  if (strcasecmp(option, "OUTPUTS") == 0) {
+    argidx++;
+    for (size_t i=0; i<argc-argidx-1; i++) {
+      name = RedisModule_StringPtrLen(argv[argidx + i], &namelen);
```
mnunberg: See above.

src/redisai.c
```diff
@@ -426,27 +418,67 @@ int RedisAI_Set_Graph_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
     return RedisModule_ReplyWithError(ctx, "ERR unsupported device");
   }
 
-  RAI_Graph *graph = NULL;
+  long long argidx = 4;
+
+  char **inputs = array_new(char*, argc - argidx);
```
mnunberg: You can use a variable-length-array here with some bounds sanity checking, unless the expected argument size can really be a lot..

e.g:
```
size_t remaining = argc - argidx;
if (remaining > 1000) { return RedisModule_ReplyWithError(ctx, "Too many arguments"); }
const char *inputs[remaining];
const char *outputs[remaining];
```

src/redisai.c
```diff
@@ -426,27 +418,67 @@ int RedisAI_Set_Graph_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
+  if (model == NULL){
+    return RedisModule_ReplyWithError(ctx, "ERR failed creating the model");
   }
 
   RedisModuleKey *key = RedisModule_OpenKey(ctx, argv[1],
       REDISMODULE_READ|REDISMODULE_WRITE);
   int type = RedisModule_KeyType(key);
   if (type != REDISMODULE_KEYTYPE_EMPTY &&
-      RedisModule_ModuleTypeGetType(key) != RedisAI_GraphType) {
+      RedisModule_ModuleTypeGetType(key) != RedisAI_ModelType) {
```
mnunberg: Also check `RedisModule_ModuleGetType(key) == REDISMODULE_KEYTYPE_MODULE`

src/redisai.c
```diff
@@ -576,49 +608,26 @@ int RedisAI_Run_Graph_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
-        return RedisModule_ReplyWithError(ctx, "ERR invalid number of inputs");
-      }
-      else {
-        return RedisModule_ReplyWithError(ctx, "ERR invalid number of outputs");
-      }
-    }
-
-    RedisModuleString** keys = RedisModule_Alloc(nitems * sizeof(RedisModuleString*));
-    RedisModuleString** names = RedisModule_Alloc(nitems * sizeof(RedisModuleString*));
+    RedisModuleString** keys = array_new(RedisModuleString*, argc - argidx);
```
mnunberg: ditto per above

lantiga:
Great review, thanks @mnunberg!

mnunberg:
I can merge this and apply the changes myself later on, if that is ok? @gkorland @K-Jo - do the command changes make sense?

lantiga:
I’ve got almost all the fixes done, sorry but I didn’t have the time to complete them.

I’ll submit them tonight.

lantiga:
Unless you need the PR to go in now, in which case I can merge, let me know

src/backends/tensorflow.c
```diff
@@ -186,88 +188,124 @@ RAI_Graph *RAI_GraphCreateTF(RAI_Backend backend, RAI_Device device,
...
   }
 
-  TF_DeleteGraph(graph->graph);
-  graph->graph = NULL;
+  TF_DeleteGraph(model->model);
+  model->model = NULL;
+
+  if (model->inputs) {
+    size_t ninputs = array_len(model->inputs);
+    for (long long i=0; i<ninputs; i++) {
```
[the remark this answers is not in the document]
lantiga: Done

src/backends/tensorflow.c
```diff
@@ -186,88 +188,124 @@ RAI_Graph *RAI_GraphCreateTF(RAI_Backend backend, RAI_Device device,
...
+      RedisModule_Free(model->inputs[i]);
+    }
+    array_free(model->inputs);
+  }
+
+  if (model->outputs) {
+    size_t noutputs = array_len(model->outputs);
+    for (long long i=0; i<noutputs; i++) {
```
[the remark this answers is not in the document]
lantiga: Done

src/backends/tensorflow.c
```diff
@@ -186,88 +188,124 @@ RAI_Graph *RAI_GraphCreateTF(RAI_Backend backend, RAI_Device device,
...
+
+  size_t noutputs = array_len(outputs);
+  char **outputs_ = array_new(char*, noutputs);
+  for (long long i=0; i<noutputs; i++) {
+    char* output = RedisModule_Alloc(strlen(outputs[i]) * sizeof(char));
+    strcpy(output, outputs[i]);
+    array_append(outputs_, output);
+  }
+
+  RAI_Model* ret = RedisModule_Alloc(sizeof(*ret));
```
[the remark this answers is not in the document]
lantiga: Done

src/redisai.c
lantiga: One question here: I'm using the `array` in utils so we can avoid passing the number of elements to the called functions. What's the actual advantage of avoiding `array` (it doesn't look like it has a lot of overhead)?

mnunberg: It's not so much a matter of overhead as it is a matter of yet one more pointer to keep track of and ensure is freed.

lantiga: I see now, thanks.
The downside is making the signature for the called functions a bit more complicated (with the added ninputs, noutputs). I'm not sure about this one.

src/backends/tensorflow.c
```diff
@@ -144,21 +144,23 @@ TF_Tensor* RAI_TFTensorFromTensor(RAI_Tensor* t){
...
+  buffer->length = modellen;
+  buffer->data = modeldef;
 
   TF_Status *status = TF_NewStatus();
 
-  TF_GraphImportGraphDef(graph, buffer, options, status);
+  TF_GraphImportGraphDef(model, buffer, options, status);
 
   if (TF_GetCode(status) != TF_OK) {
+    printf("ERR: Failed importing TF graph\n");
```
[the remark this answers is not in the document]
lantiga: Good idea. We should probably adopt this pattern in backend functions in general.

mnunberg:
I’d prefer a struct with actual error codes.. this is what we did in search, and it’s working quite nicely, if there’s more thna a single function which does this.

mnunberg:
For an uninitiated newcomer, it’s far easier to parse. If the parameters are too much, might I suggest to make the parameters as a struct instead, and just populate the struct… even though as it stands right now, the amount of parameters does not ssem to be excessive.

lantiga:
Ok, you have a good point.

lantiga:
Finishing up but still need to close, will submit tomorrow

src/backends/tensorflow.c
```diff
@@ -186,88 +188,124 @@ RAI_Graph *RAI_GraphCreateTF(RAI_Backend backend, RAI_Device device,
...
     // TODO: free memory
     return NULL;
   }
 
   TF_DeleteSessionOptions(sessionOptions);
   TF_DeleteStatus(sessionStatus);
 
-  RAI_Graph* ret = RedisModule_Alloc(sizeof(*ret));
-  ret->graph = graph;
+  size_t ninputs = array_len(inputs);
```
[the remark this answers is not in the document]
lantiga: This should be happening already in https://github.com/RedisAI/RedisAI/blob/master/src/utils/arr_rm_alloc.h
We are not including `arr.h` directly.

lantiga:
@mnunberg Changes are in, sorry if it took a while. I'll be around for another 30 mins.

lantiga:
Of note, I added returning error messages from the TF and Torch backends to the client.

mnunberg:
In the future it's better to force-push after approval as it makes it difficult to see what's changed between commits.

PR #61, opened 2019-02-22
get_deps: add -e, allow per-platform directory

mnunberg:
-e ensures that the retrieval fails at first error. Multi-platform
builds are useful in case the directory is moved across different
virtual platforms

### DIFF
--- a/get_deps.sh
+++ b/get_deps.sh
@@ -1,8 +1,15 @@
 #!/usr/bin/env bash
+set -e
 
 BASE_DIRECTORY=`pwd`
 
-DEPS_DIRECTORY=${BASE_DIRECTORY}/deps
+# Allow different deps for different platforms:
+PLATNAME=${OSTYPE}
+if [ -e /etc/debian-version ]; then
+    PLATNAME=${PLATNAME}-deb
+fi
+
+DEPS_DIRECTORY=${BASE_DIRECTORY}/deps-${PLATNAME}
 
 mkdir -p ${DEPS_DIRECTORY}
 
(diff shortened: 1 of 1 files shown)


get_deps.sh
```diff
@@ -1,8 +1,15 @@
...
 BASE_DIRECTORY=`pwd`
 
-DEPS_DIRECTORY=${BASE_DIRECTORY}/deps
+# Allow different deps for different platforms:
+PLATNAME=${OSTYPE}
+if [ -e /etc/debian-version ]; then
+    PLATNAME=${PLATNAME}-deb
+fi
+
+DEPS_DIRECTORY=${BASE_DIRECTORY}/deps-${PLATNAME}
```
lantiga: It's good to have platform specific `deps`, but you'll need to make sure the `DEPS_PATH` is set appropriately in the `Makefile`, see https://github.com/RedisAI/RedisAI/blob/master/src/Makefile#L8
(this is the cause of the failing test - it's not finding `redismodule.h` because it's looking for it in `deps`)

mnunberg:
Hi - see my other branch, I intend to CMake-ify this; it’s easier to work with and requires less logic in the shell script.

lantiga:
Good, I have worked with CMake in the past. I thought it was a no go for Redis Modules :-)
I've seen the branch now, I hadn't noticed it earlier.

At this point we could cmake-ify `get_deps.sh` too, what do you think?

mnunberg:
re get_deps.sh: It’s much quicker to just download stuff from a shell script than to use cmake’s built-in download functionality, but I’m trying to do it one step at a time. A question is how “hard” these dependencies might be — could one use dependencies from their package manager as well, or will they remain strictly vendored.

As far as CMake-ifying Redis modules, I know that I’ve converted RediSearch to CMake. There is a Makefile wrapper though, so a user can just run `make` and have it “work”, but all serious development happens using CMake.

lantiga:
Re: package managers. I'm not sure when distributions will support the C libraries of packages like TF and PyTorch, it might take a while.
Let's assume we'll provide them by downloading for the foreseeable future.

One thing that might be worth clarifying is that backends (libtensorflow, libtorch and in the future onnxruntime) will be loadable dynamically through a `AI.CONFIG` or similar command. The same holds for CPU-only or CPU+GPU versions of the same libraries.

A user might want to load libtorch CPU and libtensorflow GPU, or just libtorch, etc. In that case, it would be nice to
- avoid downloading the backends that won't be used
- download the right combination of backend+device[CPU, GPU]

lantiga:
Regarding this PR it looks good. However I propose to merge it when the work on the CMake branch is ready, so we keep CI green.

mnunberg:
Yep, I'll just bundle it together with CMake

mnunberg:
@lantiga regarding the packaging of the directories.. is the intent to have a single directory containing all the dependencies (specifically libs/) or is it also acceptable to have each dependency within its own prefix tree. Cmake can handle either just fine, but I think the latter is easier to vendor/version

lantiga:
The only advantage for having everything in deps/install/lib was for setting env vars in a stable way instead of them depending on the installed backends
```
LD_LIBRARY_PATH=/some/path/deps/install/lib redis-server ...
```
However, I'm all for separating if we provide a smart mechanism for setting env vars.

mnunberg:
also — how is this module being distributed? I’m going to assume that for the time being, it’s pretty much only being run out of the build tree?

lantiga:
Probably Pieter should chime in here @K-Jo

lantiga:
In general, I think we should create a structure that can be taken out of the build tree in the near future.

lantiga:
Closing as obsolete

PR #65, opened 2019-02-25
Build refactor

mnunberg:


lantiga:
Looks great @mnunberg!
A few requests (they should be quick):
- rebase onto master
- resolve the conflicts:
  - `test/basic_tests.py` still uses the old API
  - `Makefile` is gone, all that is needed here is add `src/err.c` to the list of source files in `CMakeLists.txt`

lantiga:
Merging!

PR #66, opened 2019-02-26
Update README.md

K-Jo:


lantiga:
Thanks! We should do the same for `load_model.sh`, `load_torch_model.sh`, `run_script.sh` in `examples/models`

Issue #64, opened 2019-02-24
Implement serialization and replication


lantiga:
So far we didn't really take care of serialization and replication for RedisAI data types.
It should be straightforward, with one caveat: in MODELSET we take the protobuf, deserialize it and throw it away (we only keep the model in memory, not the protobuf as it would be wasteful).
However, we will need to serialize the protobuf as-is (pre deserialization) for persistence and replication within the call to ModelSet.

Related to #63.

lantiga:
Closed by #72

PR #72, opened 2019-03-02
Add serialization of data types

lantiga:
This PR addresses #64.

In a nutshell:
- tensors are serialized from the underlying dlpack structures
- scripts are serialized from their string definition
- models are (re-)converted to protobufs from their in-memory representations, to avoid duplication

~~**NOTE** this is WIP, not ready for review yet (needs testing).~~


src/model.c
```diff
@@ -14,17 +14,130 @@
...
+  return model;
 }
 
-static void Model_RdbSave(RedisModuleIO *rdb, void *value) {
-  //todo
+static void RAI_Model_RdbSave(RedisModuleIO *io, void *value) {
+  RAI_Model *model = (RAI_Model*)value;
+  char *buffer = NULL;
+  size_t len = 0;
+  RAI_Error err = RAI_InitError();
```
mnunberg: I should've commented on this earlier, but zero-initialization is preferred over structure by-value returns.

lantiga: Done using zero-initialization

src/model.c
```diff
@@ -14,17 +14,130 @@
...
+    RAI_ClearError(&err);
+    if (buffer) {
+      RedisModule_Free(buffer);
+    }
+    return;
+  }
+
+  // AI.MODELSET model_key backend device [INPUTS name1 name2 ... OUTPUTS name1 name2 ...] model_blob
+
+  RedisModuleString **inputs_ = RedisModule_Calloc(sizeof(RedisModuleString*), model->ninputs);
```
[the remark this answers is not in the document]
lantiga: Good point, I'll switch to `array_`

lantiga:
@mnunberg Finally all tests check, both with AOF and RDB

I still haven't switched to `array_`, I can do that now. But otherwise feel free to review.

lantiga:
I had to temporarily disable AOF as I stumbled on something that needs further investigation: AOF loading seem to choke on commands that use `RedisModule_BlockClient`, such as `MODELRUN`.

lantiga:
Confirmed: https://github.com/antirez/redis/blob/unstable/src/aof.c#L808
We should find a way to execute all commands sequentially and not blocking if we're loading from AOF.

lantiga:
This is an issue for later, we should merge and open a separate issue.

PR #78, opened 2019-03-03
get_deps: use `deps` directory by default rather than platform specific

mnunberg:
one


K-Jo:
@mnunberg can we close this since it looks like a duplicate of #79

Issue #85, opened 2019-03-06
Add variant of MODELRUN that returns outputs instead of storing them at keys


lantiga:
This variant could be triggered by simply omitting the `OUTPUTS` arguments in the command.

lantiga:
Closing in view of #88

PR #86, opened 2019-03-07
Make OUTPUTS optional in MODELRUN and SCRIPTRUN

lantiga:
This PR addresses #85.

The `OUTPUT` section of the `MODELRUN` and `SCRIPTRUN` commands is now optional.
* If given, the commands store the output at the specified keys
* If not, the commands return the results directly without storing them

Example:
```
AI.MODELSET foo TF GPU INPUTS a b OUTPUTS mul < graph.pb

AI.TENSORSET a FLOAT 2 VALUES 2 3
AI.TENSORSET b FLOAT 2 VALUES 2 3

AI.MODELRUN foo INPUTS a b
```
returns the resulting tensor immediately, while
```
AI.MODELRUN foo INPUTS a b OUTPUTS c
```
stores the resulting tensor at key `c`


lantiga:
We might want to evaluate a variant in which we also allow inputs to be specified inline, like:
```
AI.MODELRUN foo INPUTTENSORS FLOAT 2 BLOB 03687243 FLOAT 2 BLOB 97653045
```
and
```
AI.SCRIPTRUN foo bar INPUTTENSORS FLOAT 2 BLOB 03687243 FLOAT 2 BLOB 97653045
```
Thoughts?

lantiga:
@MeirShpilraien @mnunberg I'd like to merge this one soon.
Let's not consider my last comment on having a variant with inputs inline for now.

lantiga:
Thank you @mnunberg.

The name of the command is `MODELRUN` (so among the `MODEL` commands, not `TENSOR` commands). You're suggesting something like `MODELEVAL` rather than `MODELRUN`?

To highlight the fact that it doesn't have side effects on the keys, replication, etc?

I agree Re: command parsing, I'll open a new issue.

mnunberg:
Yes, MODELEVAL, because not only does it not have side effects, the return values differ. This would make branching easier in the code, too

lantiga:
Agreed, although my primary concern is the user experience right now.

Let's check:
@gkorland @K-Jo @hhsecond what do you think about introducing `MODELEVAL`?

K-Jo:
https://redis.io/commands/sinter https://redis.io/commands/sinterstore
Fair enough to make it two commands.  Should we make them more Redis-Like as the above two commands?  We could make the one that persists something like `AI.MODELRUNSTORE`
@itamarhaber adding you for the semantics

hhsecond:
Having two commands look ambiguous and confusing. Unless, this is the way other Redis modules have been designed, we should not do that at all. From my perspective, the initial look of the PR is good enough (MODELRUN returns output if `OUTPUTS` is not specified). We could even try the way PyTorch handles in-place operations - it saves the output to `OUTPUTS` if user specifies `OUTPUTS` but regardless of that, the command returns the value always. Does that have any performance implication?

lantiga:
Thanks @hhsecond, a few considerations on these points:

- I talked to @MeirShpilraien and he explained to me the value of separating `read`-only commands from `write` commands: if you have a master-replica setting, you are allowed to run `read` commands on replicas (`MODELRUN` that returns the value would be one of them) but not `write` commands (i.e. `MODELRUN` that stores the output at key); in this sense, I also preferred to have just one command, but this is a compelling reason for separating them;

- related to returning always, there are times where you just want to run the model with inputs (e.g. an LSTM cell called sequentially) and after several evaluations retrieve the result; or when you want to run multiple models, compute an ensemble (using `SCRIPTRUN`) and retrieve that. In these cases returning always is a waste of bandwidth.

I have a proposal for naming the command that returns: `MODELRUNGET` and `SCRIPTRUNGET`. How about that? @K-Jo @mnunberg 

The rationale for the naming (vs `MODELRUNSTORE`) is that I'd like the form with the simplest name (`MODELRUN`) to be the default usage pattern, and I'd like the default usage pattern to store the result at key (it allows more flexibility).

lantiga:
See #88

PR #89, opened 2019-03-16
Small api changes:

MeirShpilraien:
1. add REDISAI_LLAPI_VERSION and RedisAI_GetAPIVersion to check api version.
2. change RAI_Backend and RAI_Device from enum to define so it will be possible to use them from the redisai.h only.

### DIFF
--- a/src/redisai.h
+++ b/src/redisai.h
@@ -4,6 +4,8 @@
 #include <stdbool.h>
 #include "redismodule.h"
 
+#define REDISAI_LLAPI_VERSION 1
+
 #define MODULE_API_FUNC(x) (*x)
 
 typedef struct RAI_Tensor RAI_Tensor;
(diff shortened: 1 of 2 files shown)


src/redisai.h
```diff
@@ -54,16 +60,18 @@ void MODULE_API_FUNC(RedisAI_ScriptRunCtxFree)(RAI_ScriptRunCtx* mctx);
...
 
 #define REDISAI_MODULE_INIT_FUNCTION(name) \
   if (RedisModule_GetApi("RedisAI_" #name, ((void **)&RedisAI_ ## name))) { \
     printf("could not initialize RedisAI_" #name "\r\n");\
-    return false; \
+    return REDISMODULE_ERR; \
   }
 
-static bool RediDL_Initialize(){
+static int RediAI_Initialize(){
```
lantiga: This should be `RedisAI_Initialize` (missing `s`)

MeirShpilraien: fixed, thanks.

lantiga:
Merged, thank you Meir!

PR #93, opened 2019-03-19
Script fixes and enhancement

itamarhaber:
@lantiga @MeirShpilraien unsure about the Ex approach, maybe just break the "API"?


lantiga:
Thank you Itamar! I think we should just drop the old API and have `RAI_ScriptRunCtxCreateEx` become the only `RAI_ScriptRunCtxCreate`. What do you think?

itamarhaber:
Makes sense - changed.

PR #95, opened 2019-03-22
Dependency map between RedisAI and Backends

hhsecond:
This information is extremely important for users especially since serialization format of some versions of PyTorch/Tensorflow might not work with other versions.


lantiga:
Thank you Sherin. In the future we will probably be more flexible but for now this is essential to put upfront.

PR #103, opened 2019-03-27
Play nice with filesystems without owner samantics

rafie:
Play nice with filesystems without owner samantics (e.g. nfs w/root_squash)

- cp --no-preserve=ownership
- tar --no-same-owner


lantiga:
Thank you Rafie

PR #111, opened 2019-04-19
[Testcases] modelrun without input and output

hhsecond:


lantiga:
This is solved by #114, as soon as that is merged, I'll also merge this one.
Only, just remove the comments:
```
# These tests fails because Rai never returns
```
Thanks!

hhsecond:
Done

Issue #116, opened 2019-05-03
MODELRUN fails to return error in case of wrong number of outputs


daron1337:
Example:
If the model expects more than one output and AI.MODELRUN is executed with only one output key, redis server logs the error but does not propagate it to the client.
The task is frozen until user manually kills it.

```
import redis
import numpy as np
import torch

class MyModule(torch.jit.ScriptModule):
   def __init__(self):
       super(MyModule, self).__init__()

   @torch.jit.script_method
   def forward(self, a, b):
       return a + b, a * b

my_script_module = MyModule()

print(my_script_module(torch.rand(2), torch.rand(2)))
my_script_module.save(“foo.pt”)

def set_redis_models():
   redis_client = redis.StrictRedis()
   print(“Redis client connected successfully”)

   with open(‘foo.pt’, ‘rb’) as f:
       model_pb = f.read()

   print(‘setting model...’)
   redis_client.execute_command(‘AI.MODELSET’, ‘model_test_cpu’, ‘TORCH’, ‘CPU’, model_pb)
   print(‘model set correctly’)
   x = np.ones((1), dtype=np.float32)
   print(‘setting tensor...’)
   redis_client.execute_command(‘AI.TENSORSET’, ‘tensor_test’, ‘FLOAT’, ‘1’, ‘BLOB’, x.tobytes())
   print(‘tensor  set correctly’)
   print(‘running model...’)
   redis_client.execute_command(‘AI.MODELRUN’, ‘model_test_cpu’, ‘INPUTS’, ‘tensor_test’, ‘tensor_test’, ‘OUTPUTS’, ‘output_test’)
   print(‘model run correctly’)

set_redis_models()
```

lantiga:
Thank you @daron1337 and @dnlcrl ! #117 should fix the issue.

lantiga:
Solved since #117 was merged

PR #117, opened 2019-05-04
Fix handling errors when they happen in the processing thread

lantiga:
This PR solves #116.

When an error was generated inside the run processing thread a) the client wasn't unblocked b) even if it were, the error message would not have been set back to the client correctly.

This PR fixes both issues.

### DIFF
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -578,6 +578,7 @@ struct RedisAI_RunInfo {
   RedisModuleString **outkeys;
   RAI_ModelRunCtx* mctx;
   int status;
+  RAI_Error* err;
 };
 
 void RedisAI_FreeRunInfo(RedisModuleCtx *ctx, struct RedisAI_RunInfo *rinfo) {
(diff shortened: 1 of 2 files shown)


lantiga:
@MeirShpilraien @mnunberg I need to merge this today, let me know if you'd still like to review. Thanks!

lantiga:
Merging, feel free to reopen.

src/redisai.c
```diff
@@ -578,6 +578,7 @@ struct RedisAI_RunInfo {
   RedisModuleString **outkeys;
   RAI_ModelRunCtx* mctx;
   int status;
+  RAI_Error* err;
```
mnunberg: Why a pointer member?

lantiga: Right, that could indeed be the object itself. However, `RAI_Error` contains strings (for the error message) that would have to be copied on the heap anyway, so in the end I felt it was more cumbersome. Let me know if you think differently.

Issue #113, opened 2019-04-23
Remove the use of moduleRegisterApi


MeirShpilraien:
Instead use the register api which was introduce in redis 5.0.4.
If the api does not exist continue without it.

lantiga:
Hi @MeirShpilraien, can you point me to an example for the right way to use the register API in other Redis modules? Thanks!

MeirShpilraien:
Actually I do not have such example yet, But I am planning to change gears and redisearch soon, I will sent you the PR's to take a look.

lantiga:
Great, thanks Meir

MeirShpilraien:
@lantiga Pls take a look at this PR: https://github.com/RedisLabsModules/RedisGears/pull/119/files
Especially the parts that use `RedisModule_ExportSharedAPI` and `RedisModule_GetSharedAPI` , those need to be used instead of `moduleRegisterApi`.

PR #118, opened 2019-05-06
Use SharedAPI

lantiga:
This PR addresses #113.
It is modeled along https://github.com/RedisLabsModules/RedisGears/pull/119

### DIFF
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -1037,7 +1037,6 @@ int RedisAI_ScriptSet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
 #define EXECUTION_PLAN_FREE_MSG 100
 
 #define REGISTER_API(name, ctx) \
-  RedisAI_ ## name = RAI_ ## name;\
   if (RedisModule_ExportSharedAPI) {\
     if (RedisModule_ExportSharedAPI(ctx, "RedisAI_" #name, RAI_ ## name) != REDISMODULE_OK) {\
       RedisModule_Log(ctx, "warning", "Could not register RedisAI_%s", #name);\
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -1040,7 +1040,7 @@ int RedisAI_ScriptSet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
   RedisAI_ ## name = RAI_ ## name;\
   if (RedisModule_ExportSharedAPI) {\
     if (RedisModule_ExportSharedAPI(ctx, "RedisAI_" #name, RAI_ ## name) != REDISMODULE_OK) {\
-      printf("could not register RedisAI_" #name "\r\n");\
+      RedisModule_Log(ctx, "warning", "Could not register RedisAI_%s", #name);\
       return REDISMODULE_ERR;\
     }\
   }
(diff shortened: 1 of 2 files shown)


src/redisai.c
```diff
@@ -1036,62 +1036,66 @@ int RedisAI_ScriptSet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
 
 #define EXECUTION_PLAN_FREE_MSG 100
 
-#define REGISTER_API(name, registerApiCallback) \
-  if(registerApiCallback("RedisAI_" #name, RAI_ ## name)){\
+#define REGISTER_API(name, ctx) \
+  RedisAI_ ## name = RAI_ ## name;\
```
MeirShpilraien: this is not really needed unless you are using your own LLAPI somewhere.

src/redisai.c
```diff
@@ -1036,62 +1036,66 @@ int RedisAI_ScriptSet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
+  if (RedisModule_ExportSharedAPI) {\
+    if (RedisModule_ExportSharedAPI(ctx, "RedisAI_" #name, RAI_ ## name) != REDISMODULE_OK) {\
       printf("could not register RedisAI_" #name "\r\n");\
```
MeirShpilraien: We can now use RedisModule_Log cause we have the ctx.

src/redisai.h
```diff
@@ -41,7 +41,7 @@ long long MODULE_API_FUNC(RedisAI_TensorDim)(RAI_Tensor* t, int dim);
 size_t MODULE_API_FUNC(RedisAI_TensorByteSize)(RAI_Tensor* t);
 char* MODULE_API_FUNC(RedisAI_TensorData)(RAI_Tensor* t);
 
-RAI_Model* MODULE_API_FUNC(RedisAI_ModelCreate)(int backend, int device,
+RAI_Model* MODULE_API_FUNC(RedisAI_ModelCreate)(RAI_Backend backend, RAI_Device device,
```
MeirShpilraien: RAI_Backend and RAI_Device are not define in this h file, other modules will not be able to compile with it. This is why it was int and we have those defines to know which value to pass:
```
#define REDISAI_BACKEND_TENSORFLOW 0
#define REDISAI_BACKEND_TORCH 1
#define REDISAI_BACKEND_ONNXRUNTIME 2

#define REDISAI_DEVICE_CPU 0
#define REDISAI_DEVICE_GPU 1
```

src/redisai.c
lantiga: That's right!

src/redisai.h
lantiga: The thing is that since now we use function pointers, using `int` here causes a warning because the function signature is not the same as its definition.
So we either switch to `int` or expose the `enums` at the `redisai.h` level.
Which one do you think is the best option?

MeirShpilraien: I am not sure why it throw warning, is it here:
```
RedisAI_ ## name = RAI_ ## name;
```

MeirShpilraien: So I just compiled it, its this like that cause the problem, I think you can just remove this line since you use directly the RAI_ functions and not the function pointers.

lantiga: Sounds good, thanks!

lantiga:
Should be all set, ready to merge.

PR #130, opened 2019-05-23
Update mobilenet.js

JoeCianflone:
Attempting to run this file I was getting an error in node `(node:14520) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of undefined` it seems that `out_data` is looking for data in a spot that it doesn't exist. So I changed it to just look in the last position and it works.


lantiga:
Thanks @JoeCianflone! This example is not in CI and it got outdated (`TENSORGET` was returning a different number of entries early on). Merging

PR #133, opened 2019-05-24
Example: refactors the js file, adds a readme file

JoeCianflone:
I did a bunch of cleanup on the example file to make it easier to use/understand. 

1. Created a `Helpers.js` file and moved things out to it
2. Easier to configure and understand what is config data
3. Used Redis Pipelines to streamline the code and make it async
4. Added some comments to explain what is going on in the file
5. Added a `readme.md` file to the example so people know how to run it
6. Updated how the content looks when the file is executed so users have a better understanding of what is going on when it runs

### DIFF
--- a/examples/js/mobilenet.js
+++ b/examples/js/mobilenet.js
@@ -91,6 +91,7 @@ function run(filenames) {
                });
       });
    });
+
 }
 
 run(Array.from(process.argv).splice(2));
--- a/examples/js/mobilenet.js
+++ b/examples/js/mobilenet.js
@@ -7,8 +7,8 @@ const Helpers = require('./Helpers');
 const config = {
    jsonClassIndex: 'imagenet_class_index.json',
    modelFile: '../models/mobilenet_v2_1.4_224_frozen.pb',
-   inputPath: 'input',
-   outputPath: 'MobilenetV2/Predictions/Reshape_1',
+   inputNode:  'input',
+   outputNode: 'MobilenetV2/Predictions/Reshape_1',
    imageHeight: 224,
    imageWidth: 224,
 };
@@ -40,7 +40,15 @@ function run(filenames) {
           * https://github.com/luin/ioredis#pipelining
           */
          redis.pipeline()
+              .call('AI.MODELSET', 'mobilenet', 'TF', 'CPU', 'INPUTS', config.inputNode, 'OUTPUTS', config.outputNode, buffer)
               .call('AI.TENSORSET', 'input_' + key, 'FLOAT', 1, config.imageWidth, config.imageHeight, 3, 'BLOB', buffer)
+              /**
+               * Important note: we're using the same input/output keys here...why? Well, in this example
+               * we're not too concerned about anything staying around, we're looking at the images
+               * checking for matches and moving along. In the real world you may need some of
+               * these keys and data to do other things, in which case, you'd want to make
+               * sure you're not overwriting all your keys :)
+               */
               .call('AI.MODELRUN', 'mobilenet', 'INPUTS', 'input_' + key, 'OUTPUTS', 'output_' + key)
               .callBuffer('AI.TENSORGET', 'output_' + key, 'BLOB')
               .exec((err, results) => {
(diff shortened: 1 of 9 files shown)


lantiga:
Great job! LGTM, merging.
Thanks so much for the contribution @JoeCianflone

examples/js/mobilenet.js
```diff
@@ -1,89 +1,89 @@
...
+         console.log("...Setting input tensor \n...running model");
+
+         /**
+          * Redis Pipelines, this allows us to set up and batch our commands
+          * If you want to learn more about how IORedis does this check out
+          * https://github.com/luin/ioredis#pipelining
+          */
+         redis.pipeline()
+              .call('AI.TENSORSET', 'input_' + key, 'FLOAT', 1, config.imageWidth, config.imageHeight, 3, 'BLOB', buffer)
+              .call('AI.MODELRUN', 'mobilenet', 'INPUTS', 'input_' + key, 'OUTPUTS', 'output_' + key)
```
lantiga: Here you're first setting the image to a key, then setting the output tensor to the same key. That's fine but maybe it's a bit counterintuitive, or the rationale could be explained in a comment.

I'm not saying it's wrong, it just needs to be explained I guess.

My first instinct for evaluating the model on a list of images and just get the result is to use the same input key and output key for all images in a sequence - if I'm just interested in the output there's no point in keeping them in different keys.

This works unless you want to keep results around, in which case you have to use different keys. At that point you might want to set and EXPIRE to those keys or take care of calling DEL on them when you don't need them anymore.

A comment on why you're using the same key for input and output and different keys for a sequence of images would be good.

examples/js/mobilenet.js
```diff
@@ -1,89 +1,89 @@
...
+const fs = require('fs');
+const Redis = require('ioredis');
+const Jimp = require('jimp');
+const Helpers = require('./Helpers');
+
+// Simple configuration, all these files should exist already in the example folder
+const config = {
+   jsonClassIndex: 'imagenet_class_index.json',
+   modelFile: '../models/mobilenet_v2_1.4_224_frozen.pb',
+   inputPath: 'input',
```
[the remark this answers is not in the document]
JoeCianflone: Ahh, I assumed they were paths because of the `/` in output, so I changed them both for consistency, `inputNode` and `outputNode` make sense to me

examples/js/mobilenet.js
JoeCianflone: Yeah. My assumption here was that this data didn't need to be kept around, but this might give someone the impression that the keys should be the same or that they're tied together in some way in which they're not, I'll adjust

JoeCianflone: ...I have no idea how I dropped that!

JoeCianflone:
I have no idea how I dropped the set, so I'll fix that

lantiga:
@JoeCianflone no pressure, just let me know if you need help with completing the PR. Thanks!

JoeCianflone:
@lantiga sorry for the delay I was out for a bit, I've addressed the issues here and I've pushed my changes. Let me know if you need anything else!

examples/js/mobilenet.js
```diff
@@ -1,89 +1,97 @@
...
+                   * work. When the pipeline is executed the output is
+                   * put into the results as part of an array, so
+                   * in this case, we don't need the output
+                   * from the first or second `.calls` we
+                   * only need the data from the
+                   * AI.TENSORGET() so this
+                   * just helps us get to
+                   * the right data
+                   */
+                  let bufferResults = results[2][1];
```
hhsecond: First of all, kudos for the comments you have made. It's so easy to follow now. Thanks a lot.
Here you might have to make the index from `2` to `3` since the `tensorset` is the 4th command in the pipeline

hhsecond:
@JoeCianflone We are cleaning up the repository and would like to remove the example folder from here and add it to our example repo. Would you like to finish the PR and then we'll take it from here or if you wish, you could create another PR to the example repo.

lantiga:
Forcing merge prior to migration of examples to the examples repo

PR #134, opened 2019-05-26
change LLAPI to use RedisModule_GetSharedAPI instead of RedisModule_GetApi

MeirShpilraien:
Because we now using RedisModule_ExportSharedAPI we need to use RedisModule_GetSharedAPI in order to get the exported api in other modules.


lantiga:
Looks great, thank you @MeirShpilraien

Issue #122, opened 2019-05-22
Cache protobuf instead of re-serializing on every RDB and AOF


lantiga:

gkorland:
@lantiga  won't it waste too much memory?

lantiga:
My concern is mainly around a model on the GPU: it doesn't make sense to me to periodically serialize it (i.e. fetch it from GPU) if it doesn't change.
To save memory we could use a tmp file and re-serialize only if it's not there, making sure the former doesn't go out of date. What do you think?

lantiga:
Maybe give the option to use cache or re-serialize

lantiga:
I suspect this is related to #302. Serialization could be triggered while the model is running in another thread. We need to switch to caching to be safe and avoid locks.
Will send a PR shortly.

Issue #33, opened 2019-02-08
Add ONNXRuntime backend


lantiga:
For more info on ONNXRuntime see https://github.com/Microsoft/onnxruntime

It has very extensive support for the ONNX standard.
It also supports ONNX-ML, allowing ML models created, say, with sklearn, to be executed in RedisAI.

PR #138, opened 2019-06-05
ONNXRuntime backend

lantiga:
This PR adds the ONNXRuntime (https://github.com/microsoft/onnxruntime) backend, allowing RedisAI to execute deep learning models in ONNX format as well as other machine learning models in ONNX-ML format. It addresses #33.

These latter models can be easily exported from scikit-learn using sklearn-onnx, see https://github.com/onnx/sklearn-onnx.

The PR uses ONNXRuntime version 0.4.0, which only supports loading a model given the filename. This forces us to write the model to a tmp file and load it back to set it in ONNXRuntime. The possibility to load it from an in-memory buffer has been added in `master`: as soon as a new version is released, we'll remove the temp file hack.

Another thing: the buffer containing the model is now cached alongside with the model itself. This is because ONNXRuntime has currently no possibility to re-serialize a model from memory to a buffer, so we need to keep the buffer around. This could change in the future, and in any case it's related to #122.


src/backends/onnxruntime.c
```diff
@@ -206,8 +237,8 @@ RAI_Model *RAI_ModelCreateORT(RAI_Backend backend, RAI_Device device,
   // loading a model from a buffer, we'll keep this as is
   char tmpfile[256];
   FILE *fp;
-  snprintf(tmpfile,256,"temp-%d.onnx", (int) getpid());
-  fp = fopen(tmpfile,"w");
+  snprintf(tmpfile, 256, "temp-%d.onnx", (int) getpid());
```
rafie: why not use mkstemp?

src/backends/onnxruntime.c
```diff
@@ -69,11 +69,17 @@ DLDataType RAI_GetDLDataTypeFromORT(ONNXTensorElementDataType dtype) {
   return (DLDataType){ .bits = 0 };
 }
 
-OrtValue* RAI_OrtValueFromTensor(RAI_Tensor* t) {
+OrtValue* RAI_OrtValueFromTensor(RAI_Tensor* t, RAI_Error *error) {
   // TODO: create outside and pass?
   OrtAllocatorInfo* allocator_info;
   OrtStatus* status;
   status = OrtCreateCpuAllocatorInfo(OrtArenaAllocator, OrtMemTypeDefault, &allocator_info);
+  if (status != NULL) {
```
rafie: It's possible to handle error with a "goto error" to an "error" label at the end of the function, instead of repeating the error block for every failure.

src/backends/onnxruntime.c
```diff
@@ -348,28 +415,41 @@ int RAI_ModelRunORT(RAI_ModelRunCtx* mctx, RAI_Error *error) {
...
+      printf("ERR: non-tensor output from ONNX models, ignoring (currently unsupported).\n");
+    }
     OrtReleaseValue(outputs[i]);
   }
 
   for (size_t i=0; i<n_input_nodes; i++) {
     OrtReleaseValue(inputs[i]);
   }
 
+  OrtReleaseAllocator(allocator);
```
rafie: Need to call OrtReleaseStatus() is successful?

src/backends/onnxruntime.c
```diff
@@ -0,0 +1,217 @@
...
+      case 64:
+        return TF_INT64; break;
+      default:
+        return 0;
+    }
+  }
+  else if (dtype.code == kDLUInt) {
+    switch (dtype.bits) {
+      case 8:
+        return TF_UINT8; break;
```
rafie: Possible to remove break.

src/backends/onnxruntime.c
```diff
@@ -0,0 +1,217 @@
...
+      case 16:
+        return TF_UINT16; break;
+      default:
+        return 0;
+    }
+  }
+  return 0;
```
rafie: Possible to remove return.

src/backends/onnxruntime.c
```diff
@@ -0,0 +1,217 @@
...
+
+  // TODO: take from
+  // https://github.com/microsoft/onnxruntime/blob/master/csharp/test/Microsoft.ML.OnnxRuntime.EndToEndTests.Capi/C_Api_Sample.cpp
+
+  OrtStatus* onnx_status;
+
+  OrtEnv* env;
+  onnx_status = OrtCreateEnv(ORT_LOGGING_LEVEL_WARNING, "test", &env);
+
+  if (onnx_status != NULL) {
```
rafie: if (onnx_status != NULL || !env) {

src/model.c
```diff
@@ -205,6 +212,9 @@ void RAI_ModelFree(RAI_Model* model, RAI_Error* err) {
   else if (model->backend == RAI_BACKEND_TORCH) {
     RAI_ModelFreeTorch(model, err);
   }
+  else if (model->backend == RAI_BACKEND_ONNXRUNTIME) {
+    RAI_ModelFreeORT(model, err);
+  }
   else {
     RAI_SetError(err, RAI_EUNSUPPORTEDBACKEND, "Unsupported backend\n");
     assert(0);
```
rafie: better to abort() as assert() does nothing if NDEBUG is defined.

src/backends/onnxruntime.c
lantiga: Since the command is executed on the main thread, I'm trying to avoid having too many temp files around. BTW, as soon as the new release for ONNXRuntime is out, we can remove this section of the code. Hopefully it'll be before GA

src/backends/onnxruntime.c
lantiga: I'm not sure why here. While it's true that `DLDataTypeCode` can only be `kDLInt`, `kDLInt`, `kDLInt`, the `DLDataType` struct has `code` as `uint8_t` (see https://github.com/dmlc/dlpack/blob/master/include/dlpack/dlpack.h#L99), so we can't exclude there's an unexpected value there.

Would you rather `abort`?

src/backends/onnxruntime.c
lantiga: In the ONNXRuntime API, `status` is only ever non-NULL if there's an issue with the function returning `status`. So if everything goes well, `status` is never allocated.

src/backends/onnxruntime.c
lantiga: Done, I just had to enclose code where variable length arrays were initialized in braces to avoid the compiler complaining about goto jumping over variable initialization.

src/model.c
lantiga: > Also, need to check pointer arguments for null values.

I've added a check for `session` that was missing. Can you maybe suggest other spots?

Issue #124, opened 2019-05-22
Jupyter notebook example to ONNXRuntime


lantiga:
http://onnx.ai/sklearn-onnx/auto_examples/plot_convert_model.html#sphx-glr-auto-examples-plot-convert-model-py [Small]

hhsecond:
@lantiga I assume this refers to the jupyter example for training, exporting and loading the model onto RedisAI rather than ONNXRuntime? If yes, this could be useful?
https://colab.research.google.com/drive/1OefFEa073bnggrO2AUbu83Fcof_hGA4K

lantiga:
Looks good! Can you make it a Jupyter Notebook and submit a PR? Thanks!

hhsecond:
@lantiga Done. It depends on the redisai-py ONNX PR. But for now, we are installing the package from specific github branch.

PR #139, opened 2019-06-11
Example with onnx

hhsecond:
Addressing #124

### DIFF
--- a/examples/onnx/ScikitLearn2Production.ipynb
+++ b/examples/onnx/ScikitLearn2Production.ipynb
@@ -1,0 +1,476 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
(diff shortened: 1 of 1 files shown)


lantiga:
On second though, this probably belongs to the `examples` repo. What do you think?

hhsecond:
That's what my instinct was too. But  then we have at least one example for each backend in this repo. If we are planning to remove all the examples from here, it make sense to move it there.

lantiga:
I agree, the examples folder needs to be cleaned up. However, some of it can go in `redisai-examples`, some as tests. Let's use this opportunity to see what goes where.

K-Jo:
I believe that both folder here and in ai-examples need a clean up.  However keeping this dangling here now we made a release feels poor.  I'll make a ticket to force us to clean this up

PR #143, opened 2019-06-13
Fixed RLEC problem of loading redisai.so

rafie:
Fixed RLEC problem of loading redisai.so by setting RUNPATH to /opt/redislabs/lib (via patchelf).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,3 +54,10 @@ IF (APPLE)
     SET_TARGET_PROPERTIES(redisai PROPERTIES
         LINK_FLAGS "-undefined dynamic_lookup")
 ENDIF()
+
+# create ramp/redisai.so with RUNPATH set to /opt/redislabs/lib for RLEC compliance
+add_custom_command(TARGET redisai POST_BUILD
+    COMMAND mkdir -p $<TARGET_FILE_DIR:redisai>/ramp
+    COMMAND cp -f $<TARGET_FILE:redisai> $<TARGET_FILE_DIR:redisai>/ramp/
+    COMMAND patchelf --set-rpath /opt/redislabs/lib $<TARGET_FILE_DIR:redisai>/ramp/`basename $<TARGET_FILE:redisai>`
+    )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,10 +54,3 @@ IF (APPLE)
     SET_TARGET_PROPERTIES(redisai PROPERTIES
         LINK_FLAGS "-undefined dynamic_lookup")
 ENDIF()
-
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -53,4 +53,11 @@ SET_TARGET_PROPERTIES(redisai PROPERTIES SUFFIX ".so")
 IF (APPLE)
     SET_TARGET_PROPERTIES(redisai PROPERTIES
         LINK_FLAGS "-undefined dynamic_lookup")
-ENDIF()
\ No newline at end of file
+ENDIF()
+
+# create ramp/redisai.so with RUNPATH set to /opt/redislabs/lib for RLEC compliance
+add_custom_command(TARGET redisai POST_BUILD
```
lantiga: Can we make these commands only run in case of a RLEC build?

Makefile
```diff
@@ -0,0 +1,53 @@
+
```
lantiga: Is the intention to build RedisAI starting from this Makefile in the future, even in non-RLEC builds?

Makefile
```diff
@@ -0,0 +1,53 @@
...
+ifeq ($(CUDA),1)
```
lantiga: So to force a non-CUDA build on a CUDA machine, we would have to set the `CUDA` env var to something different than `1`, right? We should document it somewhere.

CMakeLists.txt
rafie: The commands create a custom redisai.so that's packed into the RAMP file. I don't think there is a use case for RAMP files outside RLEC. We can do the following:
(1) Exclude the custom build for Mac,
(2) Apply patchelf as part of RAMP file creation rather than in CMake.
I think (2) is the way to go. Also, please see my comment on the Makefile.

Makefile
rafie: This Makefile is an intermediate step to the Makefile that will be introduced by the ARM branch.
The idea is twofold:
(1) Provide a convenient interface of invoking build/testing functionality,
(2) Move build logic away from .circleci/config.yml and Dockerfile into Makefiles and scripts.

So basically, packing logic introduced here in Makefile would move into its own script.
As for starting a build, I guess that one will simply invoke "make" which would then invoke cmake.

Makefile
rafie: We can offer both CUDA=1 and CPU=1 switches.
I'm planning to add some documentation via the ARM branch.

CMakeLists.txt
lantiga: I agree

Makefile
lantiga: I'm good with this strategy, especially since the arm64 build revamp is on its way.

Makefile
lantiga: Ok

PR #148, opened 2019-06-21
Fix for "failed to load: libgomp.so.1"

rafie:
"apt-get install libgomp1" added for runner.


itamarhaber:
@K-Jo this merits a new release as redisedge is broken because of that...

PR #149, opened 2019-06-25
add onnxruntime to docs get deps and cuda note

K-Jo:


gkorland:
@K-Jo we can live without CUDA the user just needs to call 
```get_deps.sh CPU``` the default is GPU

**Comment by lantiga** [comment]
@K-Jo I agree, we should point out that Linux users without a GPU should invoke
```
get_deps.sh cpu
```

**Comment by lantiga** [comment]
@K-Jo can you update with a sentence saying that for cpu only on linux the user should call
```
bash get_deps.sh cpu
```
?

This whole thing will likely change in the near future once dynamic loading is in, but it's good to merge this now.

K-Jo:
@lantiga done

Issue #47, opened 2019-02-13
Refactor multi-backend support to use function pointers


lantiga:
See https://github.com/RedisAI/RedisAI/pull/45#discussion_r256281991

lantiga:
Closed by #159

Issue #36, opened 2019-02-08
Add CONFIG command for configuring/loading backends


lantiga:
Supporting multiple backends (and multiple versions of a backend, e.g. CPU-only, CUDA etc) will have to happen on a dynamic configuration basis, using a `CONFIG` command.

For instance, if we're not using TF, there's no point in loading the libraries or the CUDA kernels.

One tricky part could be unloading, so for the first iteration we wouldn't go back after a load.

TorchScript support could be loaded by default, but we'll need to decide what happens if the default is CPU-only and we then request CUDA `libtorch` support.

So at first we would require a `CONFIG` command even for basic script support (i.e. no backend would be loaded by default).

hhsecond:
@lantiga @K-Jo What do you think the best way of distributing Docker images in this case. We can definitely support docker images for each backend separately but we should support a single image with all the backends as well (we can let the user load it dynamically but we probably need to install all the required CUDA versions beforehand).

lantiga:
@hhsecond Good point. I think having different container images for each backend is a bit overkill and in any case it could limit what users can do (the PyTorch backend is needed for SCRIPT, etc).
I'd rather build at most 3 images:
- all CPU
- TF GPU, TORCH GPU, ONNX CPU 
- all GPU (with CUDA 9.0 + 9.1)

gkorland:
@mnunberg @MeirShpilraien can you please see if you can help

lantiga:
FYI, @MeirShpilraien helped me and the issue is now solved.

andresrinivasan:
> @hhsecond Good point. I think having different container images for each backend is a bit overkill and in any case it could limit what users can do (the PyTorch backend is needed for SCRIPT, etc).
> I'd rather build at most 3 images:
> 
> * all CPU
> * TF GPU, TORCH GPU, ONNX CPU
> * all GPU (with CUDA 9.0 + 9.1)

When we add TPU and AMD GPU would that go in #2 and #3?

lantiga:
@andresrinivasan these would be specific enough to warrant separate containers IMHO

lantiga:
Closed by #159

lantiga:
Let's keep the comments in mind for packaging

PR #159, opened 2019-07-16
[WIP] Dynamic backend

lantiga:
This PR addresses #36 (and #47)

It introduces the `AI.CONFIG` command with a `LOADBACKEND` subcommand (the only one for now but there will likely be more).

Backend code in the `src/backends/` directory are compiled to their individual `.so` libraries and not loaded when the `AI` module is loaded. Instead, the user loads backends on demand by specifying the backend type (`TF`, `TORCH` or `ONNX`) and the library for the backend (which could be pointing to any CPU, GPU version thereof, provided they were compiled by linking to the backends).

To avoid pollution, we allow only one backend to be loaded once per backend type. That is, you can't have two versions of `TORCH` loaded in the same instance at the same time. This is also in relation to what happens when CUDA kernels are loaded on the GPU. Any more sophisticated behavior will have to be tested. For the same reason, unloading hasn't been implemented yet.

Backend loading works as follows:

```
r = redis.Redis(host='localhost', port=6379)

ret = r.execute_command('AI.CONFIG', 'LOADBACKEND', 'TF', 'redisai_tensorflow.so')

model_filename = 'test/test_data/graph.pb'

    model_pb = f.read()

ret = r.execute_command('AI.MODELSET', 'm', 'TF', 'CPU', 'INPUTS', 'a', 'b', 'OUTPUTS', 'mul', model_pb)

r.execute_command('AI.TENSORSET', 'a', 'FLOAT', 2, 2, 'VALUES', 2, 3, 2, 3)
r.execute_command('AI.TENSORSET', 'b', 'FLOAT', 2, 2, 'VALUES', 2, 3, 2, 3)

r.execute_command('AI.MODELRUN', 'm', 'INPUTS', 'a', 'b', 'OUTPUTS', 'c')

r.execute_command('AI.TENSORSET', 'd', 'FLOAT', 2, 2, 'VALUES', 2, 3, 2, 3)
ret = r.execute_command('AI.TENSORGET', 'd', 'VALUES')
print(ret)
```

~~**WIP**~~
~~The PR is mostly done, but I get a segfault when trying to call RedisModule_Calloc from within the dynamically loaded backend (both on macOS and Linux). I need to create a minimal repro code (e.g. based on the hello world module examples in Redis) so that it's easier to understand what's going on for anyone who'd like to help.~~
UPDATE: solved, thanks @MeirShpilraien!

Tests haven't been updated yet (so they will fail).


lantiga:
Alright, tests pass! Only the packaging fails (see below)

So, right now the build process includes a `make install` phase that will result in the following layout:
```
install/
install/redisai.so
install/backends/
install/backends/redisai_tensorflow.so
install/backends/redisai_torch.so
install/backends/redisai_onnxruntime.so
install/backends/redisai_tensorflow
install/backends/redisai_tensorflow/lib
install/backends/redisai_torch
install/backends/redisai_torch/lib
install/backends/redisai_onnxruntime
install/backends/redisai_onnxruntime/lib
```

Everything is built with `RPATH` set both on macOS and Linux, so every backend will have its own dependencies in a known location. `RPATH` is set so that the libraries are fully relocatable.

Given this layout it's easy to introduce multiple versions for each backend. We just need to choose proper names for the backends `.so` and the respective directories. I'll do this in a subsequent PR.

As I was mentioning earlier, the only missing bit for this to be merged is updating the `Makefile` and fixing the packaging: /cc @rafie I could really use your contribution here!

lantiga:
Backends can now be loaded through command-line arguments, eg
```
redis-server --loadmodule install/redisai.so TF install/backends/redisai_tensorflow.so TORCH install/backends/redisai_torch.so ONNX install/backends/redisai_onnxruntime.so
```

rafie:
Regarding RPATH/RUNPATH:
The implemented scheme looks solid.
However, the $ORIGIN trick (in Linux) will only work if the RUNPATH dtag (rather than RPATH) is set.
There is a way of making CMake do that, but it seems to keep ignoring the supplied path and replacing it with whatever it thinks best.
Therefore, we have to fallback to setting RUNPATH via post-link or via after-cmake stop.
This, running something like:
```
patchelf --set-rpath '$ORIGIN' redisai_tensorflow.so
```
Similirly, some of the backends .so should be patched similarly.
As for OSX, I need to investigate it a little to verify its behavior.

rafie:
Regarding the deployment scheme:
I see two options here. 
First option is to have an independant backend release model.
In such case, the deployment scheme will look something like:
```
install/redisai-1.2.3/redisai.so

install/tensorflow-1.12.0-variant/redisai_tensorflow.so
install/tensorflow-1.12.0-variant/libtensorflow.so
install/tensorflow-1.12.0-variant/libtensorflow_framework.so

install/libtorch-1.11.0/redisai_torch.so
install/libtorch-1.11.0/redisai_tensorflow.so
install/libtorch-1.11.0/libc10.so
install/libtorch-1.11.0/libcaffe2.so
install/libtorch-1.11.0/libcaffe2_detectron_ops.so
install/libtorch-1.11.0/libcaffe2_module_test_dynamic.so
install/libtorch-1.11.0/libfoxi.so
install/libtorch-1.11.0/libfoxi_dummy.so
install/libtorch-1.11.0/libgomp-8bba0e50.so.1
install/libtorch-1.11.0/libonnxifi.so
install/libtorch-1.11.0/libonnxifi_dummy.so
install/libtorch-1.11.0/libtorch.so.1

install/onnxruntime-0.4.0/redisai_onnxruntime.so
install/onnxruntime-0.4.0/libonnxruntime.so.0.4.0
```
From this, it's possible to see how we run into unclear build/release process, not to mention backend packaging.

The other option, RedisAI-centric release model, we place all artifacts under one versioned directory:
```
install/redisai-1.2.3/*.so
```
This will greatly simplify release and deployment, making mix-and-match backends a little harder, but not impossible.

lantiga:
Hi Rafi,
thank you for the feedback.

We can't put all .so in a single directory, because we need to be able to choose between alternative versions of, say, `libtensorflow` (CPU vs GPU, for instance).

I was thinking something in-between:

```
install/redisai-1.2.3/redisai.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-cpu.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-cpu/libtensorflow.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-cpu/libtensorflow_framework.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-gpu.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-gpu/libtensorflow.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-gpu/libtensorflow_framework.so
```

So every redisai version has its set of backends, which in turn are versioned across versions and variants.
We will maintain a strict policy where backend are compatible across patch versions (1.0.x), while we will re-release a minor version if the backend API changes (unfrequent but possible).

This layout would keep things tidy and expose just the files users are interested about (since they will need to LOADBACKEND the .so file).

rafie:
Wouldn't the following be more concise?
```
install/redisai-1.2.3/redisai.so
install/redisai-1.2.3/redisai_tensorflow-1.12.0-cpu/redisai_tensorflow.so
install/redisai-1.2.3/redisai_tensorflow-1.12.0-cpu/libtensorflow.so
install/redisai-1.2.3/redisai_tensorflow-1.12.0-cpu/libtensorflow_framework.so
```
I just wanted all shared object of a given backend (of a particular version/variant, including their dependent libraries) to reside on the same directory, to minimize complexity of RUNPATH hacking.

I was also wondering if it will be possible to introduce the patchelf post-build steps after we integrate my PR with the ARM work and build system modifications.

lantiga:
Hi @rafi, thank you.

It is fewer directories, but it would be more challenging for users to find the right backend .so file to load. I think the other way would facilitate pointing users to the right file.

By the way, with the libraries shipped by vendors (libtensorflow etc) I didn’t find a need to modify their rpath, only the rpath for the backend so (redisai_tensorflow.so).
It seems the libraries are already rpath-enabled and relocatable.

Tests pass in Linux, so I was wondering if the parchelf is a special need for the arm build or it’s general (sorry I’m not an expert in this).

Last: would it be possible for you to fix the packaging step, otherwise I won’t be able to merge.

Thank you for your help!

rafi:
off-topic, rafi != rafie

lantiga:
Ouch, sorry @rafi!

lantiga:
So @rafie, this is my final proposal
```
install/redisai-1.2.3/redisai.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-cpu/redisai_tensorflow.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-cpu/lib/libtensorflow.so
install/redisai-1.2.3/backends/redisai_tensorflow-1.12.0-cpu/lib/libtensorflow_framework.so
```

It's basically your last proposal but we place the library dependencies under `lib`, and we keep backends in a `backends` directory (which is easier to clean by wiping all its contents).

Are you ok with this?

rafie:
We can modify CMakeLists.txt the following way to replace RPATH with RUNPATH:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 3f4c59c..b6d7606 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,14 @@ SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
 SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
 SET(CMAKE_C_STANDARD 99)
 
+FUNCTION(ADD_LDFLAGS _TARGET NEW_FLAGS)
+       GET_TARGET_PROPERTY(LD_FLAGS ${_TARGET} LINK_FLAGS)
+       IF(LD_FLAGS)
+               SET(NEW_FLAGS "${LD_FLAGS} ${NEW_FLAGS}")
+       ENDIF()
+       SET_TARGET_PROPERTIES(${_TARGET} PROPERTIES LINK_FLAGS ${NEW_FLAGS})
+ENDFUNCTION()
+
 FIND_LIBRARY(TF_LIBRARIES NAMES tensorflow libtensorflow.so
              PATHS ${depsAbs}/libtensorflow/lib)
 IF (NOT TF_LIBRARIES)
@@ -69,6 +77,7 @@ SET_TARGET_PROPERTIES(redisai_tensorflow PROPERTIES SUFFIX ".so")
 IF (APPLE)
   SET_TARGET_PROPERTIES(redisai_tensorflow PROPERTIES INSTALL_RPATH "@loader_path/lib")
 ELSE ()
+  ADD_LDFLAGS(redisai_tensorflow "-Wl,--enable-new-dtags")
   SET_TARGET_PROPERTIES(redisai_tensorflow PROPERTIES INSTALL_RPATH "\$ORIGIN/lib")
 ENDIF()
 INSTALL(TARGETS redisai_tensorflow LIBRARY DESTINATION backends/redisai_tensorflow)
@@ -81,6 +90,7 @@ SET_TARGET_PROPERTIES(redisai_torch PROPERTIES SUFFIX ".so")
 IF (APPLE)
   SET_TARGET_PROPERTIES(redisai_torch PROPERTIES INSTALL_RPATH "@loader_path/lib")
 ELSE ()
+  ADD_LDFLAGS(redisai_torch "-Wl,--enable-new-dtags")
   SET_TARGET_PROPERTIES(redisai_torch PROPERTIES INSTALL_RPATH "\$ORIGIN/lib")
 ENDIF()
 INSTALL(TARGETS redisai_torch LIBRARY DESTINATION backends/redisai_torch)
@@ -93,6 +103,7 @@ SET_TARGET_PROPERTIES(redisai_onnxruntime PROPERTIES SUFFIX ".so")
 IF (APPLE)
   SET_TARGET_PROPERTIES(redisai_onnxruntime PROPERTIES INSTALL_RPATH "@loader_path/lib")
 ELSE ()
+  ADD_LDFLAGS(redisai_onnxruntime "-Wl,--enable-new-dtags")
   SET_TARGET_PROPERTIES(redisai_onnxruntime PROPERTIES INSTALL_RPATH "\$ORIGIN/lib")
 ENDIF()
 INSTALL(TARGETS redisai_onnxruntime LIBRARY DESTINATION backends/redisai_onnxruntime)
```

lantiga:
@rafie Done

Issue #160, opened 2019-07-18
AI.MODELRUN and AI.SCRIPTRUN are not declaring keys properly


MeirShpilraien:
The only declared key is the model/script (and not the tensors). Also there is not early exit in the code to not actually execute the command:
```
  if (RedisModule_IsKeysPositionRequest(ctx)) {
    RedisModule_KeyAtPos(ctx, 1);
  }
```

lantiga:
Addressed by #161, closing

PR #161, opened 2019-07-18
Fix use of key pos API, and point to first tensor key in command def

lantiga:
Addresses #160 
/cc @MeirShpilraien

### DIFF
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -688,13 +688,21 @@ int RedisAI_ModelRun_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
   // be picked up on the next round. We also need to signal when it's time to dispose
   // of the old model.
   // The key is having a single thread looping for execution
+  if (argc < 3) return RedisModule_WrongArity(ctx);
+
   if (RedisModule_IsKeysPositionRequest(ctx)) {
     RedisModule_KeyAtPos(ctx, 1);
+    for (int i=1; i<argc; i++) {
+      const char* arg = RedisModule_StringPtrLen(argv[i], NULL);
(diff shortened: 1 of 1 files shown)


src/redisai.c
```diff
@@ -688,14 +688,22 @@ int RedisAI_ModelRun_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
   // be picked up on the next round. We also need to signal when it's time to dispose
   // of the old model.
   // The key is having a single thread looping for execution
+  if (argc < 3) return RedisModule_WrongArity(ctx);
+
   if (RedisModule_IsKeysPositionRequest(ctx)) {
     RedisModule_KeyAtPos(ctx, 1);
+    for (int i=1; i<argc; i++) {
```
MeirShpilraien: shouldn't you start from 2?

src/redisai.c
```diff
@@ -848,14 +856,22 @@ int RedisAI_StartRunThread() {
 
 // script key, fnname, INPUTS, key1, key2 ... OUTPUTS, key1, key2 ...
 int RedisAI_ScriptRun_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+  if (argc < 4) return RedisModule_WrongArity(ctx);
+
   if (RedisModule_IsKeysPositionRequest(ctx)) {
     RedisModule_KeyAtPos(ctx, 1);
+    for (int i=2; i<argc; i++) {
```
MeirShpilraien: shouldn't you start from 3?

src/redisai.c
lantiga: Damn

lantiga:
Fixed, good catch :-)

Issue #158, opened 2019-07-10
Add AI.MODELDEL and AI.SCRIPTDEL


K-Jo:
`AI.MODELDEL model_key`
`AI.SCRIPTDEL model_key`

lantiga:
Solved by #163

PR #163, opened 2019-07-18
Implement MODELDEL and SCRIPTDEL

lantiga:
Addresses #158 
/cc @K-Jo @MeirShpilraien

### DIFF
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -592,7 +592,7 @@ int RedisAI_ModelDel_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
   RedisModuleString* keystr;
   AC_GetRString(&ac, &keystr, 0);
 
-  RedisModuleKey *key = RedisModule_OpenKey(ctx, keystr, REDISMODULE_READ);
+  RedisModuleKey *key = RedisModule_OpenKey(ctx, keystr, REDISMODULE_WRITE);
   int type = RedisModule_KeyType(key);
   if (type == REDISMODULE_KEYTYPE_EMPTY) {
     RedisModule_CloseKey(key);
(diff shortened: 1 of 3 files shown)


K-Jo:
@lantiga needs docs

test/basic_tests.py
```diff
@@ -99,6 +99,22 @@ def test_set_tensor(env):
...
+        model_pb = f.read()
+
+    con = env
+    ret = con.execute_command('AI.MODELSET', 'm', 'TF', 'CPU',
+                              'INPUTS', 'a', 'b', 'OUTPUTS', 'mul', model_pb)
+    con.assertEqual(ret, b'OK')
+
+    ret = con.execute_command('AI.MODELDEL', 'm')
+    con.assertEqual(ret, b'OK')
+
```
MeirShpilraien: I would have checked here if after the deletion the key does not exists

test/basic_tests.py
```diff
@@ -580,6 +596,20 @@ def test_set_correct_script(env):
...
+    with open(script_filename, 'rb') as f:
+        script = f.read()
+
+    ret = env.execute_command('AI.SCRIPTSET', 'ket', 'CPU', script)
+    env.assertEqual(ret, b'OK')
+
+    ret = env.execute_command('AI.SCRIPTDEL', 'ket')
+    env.assertEqual(ret, b'OK')
+
```
MeirShpilraien: Same comment here

src/redisai.c
```diff
@@ -1028,6 +1058,36 @@ int RedisAI_ScriptGet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
+
+  RedisModule_AutoMemory(ctx);
+
+  ArgsCursor ac;
+  ArgsCursor_InitRString(&ac, argv+1, argc-1);
+
+  RedisModuleString* keystr;
+  AC_GetRString(&ac, &keystr, 0);
+
+  RedisModuleKey *key = RedisModule_OpenKey(ctx, keystr, REDISMODULE_READ);
```
MeirShpilraien: Isn't this should be REDISMODULE_WRITE?

src/redisai.c
```diff
@@ -580,6 +580,36 @@ int RedisAI_ModelGet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
...
```
MeirShpilraien: REDISMODULE_WRITE?

src/redisai.c
lantiga: Good catch, thanks!

lantiga:
Thanks @K-Jo and @MeirShpilraien, I addressed both comments

PR #167, opened 2019-07-27
Build: fixed rlec_runpath_fix and packing mechanisms

rafie:


Makefile
```diff
@@ -18,15 +18,19 @@ endif
...
+
+build:
 ifeq ($(wildcard build/.),)
 	mkdir -p build
 	cd build; \
 	cmake -DDEPS_PATH=../deps/install ..
 endif
 	$(MAKE) -C build
+	@echo Fixing RLEC RUNPATH...
+	@patchelf --set-rpath '$$ORIGIN:$(REDIS_ENT_LIB_PATH)' $(BINDIR)/redisai.so
```
lantiga: This fails on macOS. We should condition this with `ifeq ($(UNAME_S),Linux)` or equivalent.

Makefile
```diff
@@ -80,4 +80,5 @@ endif
 	@set -e ;\
 	cd test ;\
 	. venv/bin/activate ;\
+	LD_LIBRARY_PATH=$(PWD)/deps/install/lib \
```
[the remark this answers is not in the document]
rafie: Well, it goes like this:
First, RPATH (which we don't set) takes precedence of LD_LIBRARY_PATH.
RUNPATH, on the other hand, gives way for LD_LIBRARY_PATH, which is useful here, because we set redisai.so RUNPATH to $ORIGIN:/opt/redislabs/lib, and since the former is at RedisAI/build/ while the engines libraries are ar RedisAI/deps/install/lib, it will not find them without help.
For that scheme to work with RLTest, I had to fix it not to completely override LD_LIBRARY_PATH with what it thinks best, but rather prepend to it.
Another tricky caveat that might hurt us is the way RPATH may be set in engine libraries (e.g. onnxruntime). If indeed RPATH is set, $ORIGIN is ignored, which may lead to failure. So we need to fix that as well. 
I'm currently validating this scheme in RLEC.

Makefile
rafie: True!
I will exclude OSX at the moment, and cover that later.

lantiga:
@rafie We'll need to re-evaluate this one. Should we close this in the meantime?

Issue #172, opened 2019-08-04
Load default backends lazily


lantiga:
In the initial design of dynamic loading of backends, RedisAI would return an error to the client if the user tried to load a model for which a backend was not loaded.

Apart from probably not being beginner-friendly, this posed an issue when loading an RDB file, since Redis would refuse to start unless the right backends were loaded as command-line arguments.

To circumvent this issue, we will have RedisAI lazily load a missing backend from the default backend, if not previously set. This occurrence will be logged by RedisAI. Default backends are named `<redisai_<backend>/redisai_<backend>.so`, i.e.

```
redisai_tensorflow/redisai_tensorflow.so
redisai_torch/redisai_torch.so
redisai_onnxruntime/redisai_onnxruntime.so
```

The idea is that non-default backends will have version specifiers in the name, and that default backends can be eventually symlinks to appropriate backends.

K-Jo:
@lantiga let's track here the work for the ramp file as well

lantiga:
@K-Jo just for consistency, can we close this and eventually track the RAMP work in a separate issue?

K-Jo:
I think it's done @rafie ? otherwise yes

PR #173, opened 2019-08-04
Load default backend lazily if not set

lantiga:
Addresses #172 

This PR introduces a new CONFIG subcommand

```
AI.CONFIG BACKENDSPATH <path>
```

which also works as a command-line argument just like the other CONFIG subcommands.

BACKENDSPATH instructs RedisAI where to look for the default backends.

In addition, in case a path for a backend file is specified as a relative path, e.g.
```
AI.CONFIG LOADBACKEND TORCH redisai_torch/redisai_torch.so
```
the given path is interpreted as relative to the BACKENDSPATH.

In this case, if BACKENDSPATH is not specified, it will be assumed to be
```
<location_of_redisai.so>/backends
```


K-Jo:
@lantiga is there a default BACKENDSPATH or will there be an exception if the command was not called or it was not set at module loadtime?

lantiga:
It defaults to <where_redisai.so_is>/backends

rafie:
Looks great. PS do we want to support env var REDISAI_BACKENDS_PATH?

lantiga:
> Looks great. PS do we want to support env var REDISAI_BACKENDS_PATH?

Right now there's a `BACKENDSPATH` command-line argument, we can surely support the env var as well. Let's merge this, can you please follow-up regarding this point at #172?

PR #174, opened 2019-08-05
Fixed error in BACKENDSPATH command

rafie:
Off by one in argv.


lantiga:
Doh! Thank you @rafie

PR #180, opened 2019-08-15
Add support for ONNXRuntime 0.5

lantiga:
Addresses #179 

Release notes for ONNXRuntime 0.5 are here: https://github.com/microsoft/onnxruntime/releases/tag/v0.5.0

Notable changes:
- we don't need to write the model to a tmp file and then load, but we load the model directly from a buffer
- support for CUDA 10 (at this point we should consider switching all backends to CUDA 10, I'll open and issue for that)
- there's initial support for OpenVINO executors (which in turn can control devices like the Intel Movidius, etc, see https://docs.openvinotoolkit.org)

### DIFF
--- a/src/backends/onnxruntime.c
+++ b/src/backends/onnxruntime.c
@@ -113,6 +113,10 @@ error:
 RAI_Tensor* RAI_TensorCreateFromOrtValue(OrtValue* v, RAI_Error *error) {
   OrtStatus* status = NULL;
 
+  RAI_Tensor* ret = NULL;
+  int64_t *shape = NULL;
+  int64_t *strides = NULL;
+ 
   int is_tensor;
   status = OrtIsTensor(v, &is_tensor);
   if (status != NULL) goto error;
@@ -123,7 +127,7 @@ RAI_Tensor* RAI_TensorCreateFromOrtValue(OrtValue* v, RAI_Error *error) {
     return NULL;
   }
 
-  RAI_Tensor* ret = RedisModule_Calloc(1, sizeof(*ret));
+  ret = RedisModule_Calloc(1, sizeof(*ret));
 
   DLContext ctx = (DLContext){
       .device_type = kDLCPU,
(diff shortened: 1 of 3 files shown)


src/backends/onnxruntime.c
```diff
@@ -115,7 +111,13 @@ OrtValue* RAI_OrtValueFromTensor(RAI_Tensor* t, RAI_Error *error) {
...
 
 RAI_Tensor* RAI_TensorCreateFromOrtValue(OrtValue* v, RAI_Error *error) {
-  if (!OrtIsTensor(v)) {
+  OrtStatus* status = NULL;
+
+  int is_tensor;
+  status = OrtIsTensor(v, &is_tensor);
+  if (status != NULL) goto error;
+
+  if (!is_tensor) {
```
rafie: I was wondering whether we should free ret, shape, and strides at the error handler to avoid leaks.

src/backends/onnxruntime.c
```diff
@@ -225,9 +237,15 @@ RAI_Model *RAI_ModelCreateORT(RAI_Backend backend, RAI_Device device,
   }
 
   // TODO: probably these options could be configured at the AI.CONFIG level
-  OrtSessionOptions* session_options = OrtCreateSessionOptions();
-  OrtSetSessionThreadPoolSize(session_options, 1);
-  OrtSetSessionGraphOptimizationLevel(session_options, 1);
+  OrtSessionOptions* session_options;
+  status = OrtCreateSessionOptions(&session_options);
+  if (status != NULL) goto error;
```
rafie: Do we need to free env if we fail here?

lantiga: Yep, and session_options a bit down. Fixed.

src/backends/onnxruntime.c
lantiga: Done

src/backends/onnxruntime.c
lantiga: Sorry, taking it back. `env` is global, it is only allocated the first time and it shouldn't be freed.

Issue #71, opened 2019-03-02
Create a queue per device, both for models and script


lantiga:
We should allow models and scripts to be placed on the available devices (CPU, GPU0, GPU1, etc) and execute concurrently.

To do so, instead of a single execution queue, we should have as many queues as we have devices.

lantiga:
This goes in tandem with the new DAG command (to enable parallel invocations from a single client).
We should enable the queue for script, but keep models and scripts on the same queue for fairness, at least at the beginning.

lantiga:
Solved by #186

PR #186, opened 2019-08-25
Queue system refactor

lantiga:
This PR introduces a refactoring of the run queue system. Specifically, now
- there's a way to optionally specify the `deviceid` (using `GPU:0`, `GPU:1`, etc) as part of model set
- one queue is instantiated per device (i.e. runs submitted to different devices will not interfere), instantiated lazily when the device is first used as part of MODELSET or SCRIPTSET (addresses #71)
- one polling thread per device, started lazily when the device is first used as part of MODELSET or SCRIPTSET
- scripts are now non-blocking, i.e. they behave like models (runs are put on their respective device queue)

Future improvements (for future PRs):
- move from polling to threading signals
- potentially allow more than one queue per device (to be handled with care)

Since this PR introduces a backward incompatible change to RDB serialization for models and scripts, this needs to be merged before GA.


src/model.c
```diff
@@ -136,9 +138,32 @@ static void RAI_Model_AofRewrite(RedisModuleIO *aof, RedisModuleString *key, voi
     array_append(outputs_, RedisModule_CreateString(ctx, model->outputs[i], strlen(model->outputs[i])));
   }
 
+  char backend[256];
```
rafie: char backend[256] = "";
to prevent catastrophe if model->backend isn't matched.

lantiga: Done

Issue #120, opened 2019-05-14
GPU Docker failed to start


srijiths:
Hi,

I have used Dockerfile-gpu to build a GPU docker. Docker build is successful, but Redis exiting with libcuda exception.
```
1:M 14 May 09:14:25.215 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
1:M 14 May 09:14:25.215 # Server initialized
1:M 14 May 09:14:25.215 # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
1:M 14 May 09:14:25.215 # WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.
1:M 14 May 09:14:25.216 # Module /usr/lib/redis/modules/redisai.so failed to load: libcuda.so.1: cannot open shared object file: No such file or directory
1:M 14 May 09:14:25.216 # Can't load module from /usr/lib/redis/modules/redisai.so: server aborting
```
Thank You,
Srijith

lantiga:
Sorry @srijiths for the late feedback. Did you run this with `nvidia-docker`?

srijiths:
@lantiga Executed this command. nvidia-docker run -it redisai/redisai:gpu

In Dockerfile-gpu FROM changed to nvidia/cuda:10.0-cudnn7-devel-ubuntu16.04 as i have Cuda 10.0.
Please find the docker build log here. [https://gist.github.com/srijiths/3db165c51ff2d6a5130958e5c73489d9](url) . Towards the end of the log , doubtful about below statements.
```
-- Found CUDA: /usr/local/cuda (found suitable version "10.0", minimum required is "7.0")
-- Caffe2: CUDA detected: 10.0
-- Caffe2: CUDA nvcc is: /usr/local/cuda/bin/nvcc
-- Caffe2: CUDA toolkit directory: /usr/local/cuda
-- Caffe2: Header version is: 10.0
-- Found CUDNN: /usr/include
[...]

PR #189, opened 2019-08-26
Docker GPU

hhsecond:
Fixing #120

### DIFF
--- a/get_deps.sh
+++ b/get_deps.sh
@@ -172,7 +172,7 @@ if [[ ! -d ${ORT_PREFIX} ]]; then
   mkdir -p ${ORT_PREFIX}
 
   if [ ! -e ${ORT_ARCHIVE} ]; then
-    echo "Downloading ONNXRuntime ${ORT_VERSION} ${DEVICE}"
+    echo "Downloading ONNXRuntime ${ORT_VERSION} ${ORT_BUILD}"
     wget -q https://github.com/microsoft/onnxruntime/releases/download/v${ORT_VERSION}/${ORT_ARCHIVE}
   fi
   
--- a/Dockerfile-gpu
+++ b/Dockerfile-gpu
@@ -8,12 +8,9 @@ RUN apt update && apt -y upgrade
 RUN apt-get install -y libgomp1
 RUN apt-get install -y wget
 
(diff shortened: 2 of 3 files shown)


hhsecond:
Few points to note here
- For making the server accessible outside, We needed `--protected-mode no`. Not sure if there is a better way
- We need to push this to redisai docker hub for using it in the CI. I have pushed one to [tensorwerk's account](https://hub.docker.com/r/tensorwerk/redisai-gpu). Once this PR is good to go, it would be great if somebody who has access could help me pushing this to redisai docker hub

get_deps.sh
```diff
@@ -5,21 +5,6 @@ HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 set -e
 [[ $VERBOSE == 1 ]] && set -x
 
-if [[ "$1" == "cpu" ]]; then
```
rafie: Let's not remove this for the time being, until gen_deps.sh is stabilized (there's quite a few changes coming up with the ARM PR).

hhsecond: Done

lantiga:
Regarding protected mode, I think (but I might be wrong @rafie) the way to go would be to not specify it.
Instead, one could mount redis.conf from the host or a docker volume with the appropriate settings (eg auth, ...).
What do you think?

hhsecond:
I think we should have `protected mode no` by default and let the user override that by mounting a config (I assume this is what base Redis image and RediAI CPU image does). I might have to move `protected mode no` from the command to the config to make it overridable

lantiga:
Thank you Sherin!

hhsecond:
This is what `Redis` does. It removes the config from the source before building `Redis` itself
https://github.com/docker-library/redis/blob/0b2910f292fa6ac32318cb2acc84355b11aa8a7a/5.0/Dockerfile#L66

Instead what I did is to handle that using the config file which can be controlled by mounting the configuration file. Also, the installation of Redis in the old `Dockercompose-GPU` was from ubuntu repo and there were few dependency issues. I had to install it from the `stable` URL. Hope that's OK

lantiga:
Looks good to me. I'm good to merge, @rafie?

PR #190, opened 2019-08-26
AI.MODELSET example TF would use a .pb file

andresrinivasan:


lantiga:
Nice catch @andresrinivasan, can you please sign the CLA so we can merge? Thanks!

lantiga:
Thanks!

PR #191, opened 2019-08-27
ARM support + os-arch-variant build architecture

rafie:


lantiga:
Hi @rafie, I tested this on my system and it worked flawlessly.

I also tested an out-of-source build, in a `RedisAI-build` directory parallel to the `RedisAI` source directory. It worked without issues, the only nit is that the `install-cpu` directory was created inside the source directory.

This was kind of unexpected, as I expected the install directory to be outside the source dir, but in fact placing it in the `RedisAI-build` directory is quite confusing too, since we'd have `redisai.so` in `RedisAI-build` and another `redisai.so` in `RedisAI-build/install-cpu`.

The ideal would be to have a `RedisAI-install` parallel to `RedisAI-build`, but we can't just create directories outside the build root.

So, long story short, we should keep it in the source tree for now :-D


I'm good to merge, but maybe we want to take this opportunity to rearrange directories as we discussed on Slack?

PR #211, opened 2019-09-08
CircleCI: multiarch docker build + selective engines build

rafie:
- Configuration for multi-arch docker build in CircleCI (introduced but disabled)
- Allow disable building specific engines (uses Filipe's code from cmake_pref branch)


get_deps.sh
```diff
@@ -74,100 +77,108 @@ fi
...
-		PT_BUILD=cpu
-	fi
 
-	[[ "$PT_VERSION" == "latest" ]] && PT_BUILD=nightly/${PT_BUILD}
+		[[ "$PT_VERSION" == "latest" ]] && PT_BUILD=nightly/${PT_BUILD}
 
-	LIBTORCH_ARCHIVE=libtorch-${PT_BUILD}-${PT_OS}-${PT_ARCH}-${PT_VERSION}.tar.gz
-	LIBTORCH_URL=https://s3.amazonaws.com/redismodules/pytorch/$LIBTORCH_ARCHIVE
+		LIBTORCH_ARCHIVE=libtorch-${PT_BUILD}-${PT_OS}-${PT_ARCH}-${PT_VERSION}.tar.gz
+		LIBTORCH_URL=https://s3.amazonaws.com/redismodules/pytorch/$LIBTORCH_ARCHIVE
```
lantiga: It was the same way earlier, but I notice there's no codepath that downloads LIBTORCH from the official PyTorch packages for x86, like
https://download.pytorch.org/libtorch/cpu/libtorch-macos-1.2.0.zip
or
https://download.pytorch.org/libtorch/cu100/libtorch-shared-with-deps-1.2.0.zip
https://download.pytorch.org/libtorch/cpu/libtorch-shared-with-deps-1.2.0.zip

I think we should do it as we do for TensorFlow and ONNXRuntime

rafie: Fixed.

lantiga:
Approved pending Docker test issues

Issue #155, opened 2019-07-02
RedisAI doesn't throw error/warning if user tries to set GPU when GPU is not available


hhsecond:
Other than PyTorch, none of the backends throws an error or a warning if the user tries to set the model with GPU when GPU is not available. There are #TODOs in the source to fix this already. Keeping the issue open to track it for the GA perhaps, because it could cause serious skepticism in certain situations.

hhsecond:
ONNXRuntime still needs to do this check dynamically but since the dynamic allocation of CUDA is decided to do in the future, this chek can also be implemented then

lantiga:
I think since we can load backends dynamically and we can build ONNXRuntime with multiple backends (CPU, GPU, TensorRT, OpenVINO, etc), we can eventually leave the dynamicity for when the user chooses what backend to load, rather than within ONNXRuntime.

lantiga:
We can have several ONNXRuntime backends, one per executor. This way they'll also be smaller and with fewer dependencies.

hhsecond:
Make sense. In that case, I should leave this issue opened since the CUDA check won't happen for ORT if the default installation is CPU

lantiga:
Are you saying that if ORT is built for CPU and you request GPU we won't return an error, right?

hhsecond:
No, if ORT is built with CPU, then we will return an error if user request for a GPU execution.

But if RedisAI is built without DEVICE var (or DEVICE=cpu) it throws an error if user request GPU even if the loaded ORT backend is running on GPU. We have problem for the other case as well -  RedisAI is built for GPU and loaded ORT backend is CPU only

 https://github.com/RedisAI/RedisAI/blob/19d15c42779423cccd7830aefc971761f9dd7e7f/src/backends/onnxruntime.c#L269-L279

lantiga:
Say we have two ORT backends, like `redisai_onnxruntime_cpu.so` and `redisai_onnxruntime_gpu.so`
- the moment we load `redisai_onnxruntime_gpu.so` it will throw an error if there's no GPU or CUDA on the machine
- if we load `redisai_onnxruntime_cpu.so` and request GPU, it returns an error (which is correct)

In other words, it doesn't really matter how you build `redisai.so`, only the backends counts. Does this make sense or I'm being overly dense? :-)

hhsecond:
Thanks for the clarification. Closing the issue

Issue #152, opened 2019-06-28
Descriptive error messages


hhsecond:
Error messages could divulge more information such as the expected number of outputs in case the number of input passed by the user is wrong. For this case, RedisAI now says `ERR: Unexpected number of outputs for graph`. But there could be other cases as well. I can take this up something next week.

PR #212, opened 2019-09-09
Cuda availability check and error messages

hhsecond:
Quick fix for #152 & #155 (Pending the GPU check for ONNXRuntime)


src/backends/onnxruntime.c
```diff
@@ -270,6 +270,12 @@ RAI_Model *RAI_ModelCreateORT(RAI_Backend backend, RAI_Device device, int64_t de
   if (device == RAI_DEVICE_GPU) {
     OrtSessionOptionsAppendExecutionProvider_CUDA(session_options, deviceid);
   }
+#else
```
hhsecond: `RAI_ONNXRUNTIME_USE_CUDA` doesn't seem to propagate with `-DDEVICE=gpu` in cmake

hhsecond: Rearranged the CMAKE file. It doesn't seem to break anything else. Please let me know

hhsecond:
Rebased and force pushed

hhsecond:
Tests are failing by the rltest bug which is being tracked here https://github.com/RedisLabsModules/RLTest/issues/52

src/backends/onnxruntime.c
hhsecond: Changed the order in CMAKE file. It should not break anything else but appreciate if you could take a look

lantiga:
Hi Sherin, can you confirm tests are passing locally on a GPU machine?

hhsecond:
@lantiga Tests should pass ideally. But after the recent changes in RLTest, all our existing tests are failing -> `con.execute_command` is not returning bytes strings anymore. I'll fix our tests soon and then we are good to go

hhsecond:
Changes in python Redis client has been reverted back and Circle tests are not failing because of those changes. But there is still one issue with load backend. Let me figure that out and then we can merge

hhsecond:
@lantiga Everything looks good in the CI

lantiga:
Great, thanks!

PR #220, opened 2019-09-12
[fix] fixed lack of deny-oom strflag on RedisAI Commands that may use additional memory

filipecosta90:
- [fix] fixed lack of deny-oom strflag on RedisModule_CreateCommand for AI.TENSORSET, AI.MODELSET, AI.MODELRUN, AI.SCRIPTSET, AI.SCRIPTRUN

### DIFF
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -1547,7 +1547,7 @@ int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) 
     return REDISMODULE_ERR;
   }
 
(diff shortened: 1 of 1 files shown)


filipecosta90:
> Looks good, just a note.
> The `deny-oom` flag here works for tensors, allocated through tensorset or as a result of the execution of modelrun or scriptrun, but the execution of models will generate intermediate tensors that are not allocated by the Redis allocator, but by whatever allocator is used in the backends (which may act on main memory or GPU memory, depending on the device).
> Memory allocation in this case would go undetected, maybe it's worth pointing this out in a comment?

Agree with you. Will create an issue for that.

PR #230, opened 2019-09-27
Jobs on GPU

hhsecond:
- CircleCI config for running GPU jobs
- Test cases running on GPU Device

I am not using any of the existing build utilities such as system-setup.py since relying on a file inside the repo trigger rebuilding of that layer on each push which is time-consuming. We need to iterate over this and other build processes to unify the build processes and to do the packaging and release of GPU docker image etc

Also, because of the `allow growth` feature in TF, the test cases on GPU takes a lot of time now

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,14 +68,12 @@ jobs:
     steps:
       - ci_steps:
           platform: debian
-
   # build-macos:
   #   macos:
   #     xcode: 10.2.1
   #   steps:
   #     - ci_steps:
   #         platform: macosx
-
   build-multiarch-docker:
     machine:
       enabled: true
@@ -100,6 +98,22 @@ jobs:
             sudo docker login -u redisfab -p $DOCKER_REDISFAB_PWD
             make -C opt/build/docker build
             sudo make -C opt/build/docker publish
+  build-gpu:
+    machine:
+      docker_layer_caching: true
+      resource_class: gpu.nvidia.medium
+      image: ubuntu-1604-cuda-10.1:201909-23
+    steps:
+      - checkout
+      - run:
+          name: Bulding Docker
+          command: |
+            docker build -f Dockerfile.gpu-test -t redisai-gpu-test .
+      - run:
+          name: Test
+          command: |
+            mkdir -p ~/workspace/tests
+            docker run --gpus all -v $HOME/workspace/tests:/data/tests/logs -it --rm redisai-gpu-test
 
   deploy_package:
     parameters:
@@ -146,6 +160,16 @@ workflows:
       #        only: /.*/
       #      branches:
       #        only: master
+      - build-gpu:
+          filters:
+            tags:
+              only: /.*/
+      - build-multiarch-docker:
+          filters:
+            tags:
+              only: /.*/
+            branches:
+              only: master
       - deploy_package:
           name: deploy_branch
           package: branch
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,20 +100,28 @@ jobs:
             sudo make -C opt/build/docker publish
   build-gpu:
     machine:
+      enabled: true
       docker_layer_caching: true
       resource_class: gpu.nvidia.small
       image: ubuntu-1604-cuda-10.1:201909-23
     steps:
       - checkout
+#      - run:
+#          name: Checkout LFS
+#          command: |
+#            curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
+#            sudo apt-get install -y git-lfs
+#            git lfs install
+#            git lfs pull
       - run:
-          name: Building Docker
+          name: Build
           command: |
-            docker build -f Dockerfile.gpu-test -t redisai-gpu-test .
+            docker build -f Dockerfile.gpu-test -t redisai-gpu:latest-x64-bionic-test .
       - run:
           name: Test
           command: |
             mkdir -p ~/workspace/tests
-            docker run --gpus all -v $HOME/workspace/tests:/redisai/test/logs -it --rm redisai-gpu-test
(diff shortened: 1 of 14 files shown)


test/basic_tests.py
```diff
@@ -13,13 +11,15 @@
 import sys
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../opt/readies"))
-import paella
```
rafie: Better to keep it. It is useful for debugging purposes (enables easily debugger invocation and insertion of breakpoints in code).

opt/Makefile
```diff
@@ -1,5 +1,5 @@
 
-ROOT:=..
+ROOT:=$(abspath ..)
```
rafie: No need for abspath here as readies/mk/main will abspath it.

Dockerfile.gpu-test
```diff
@@ -0,0 +1,65 @@
+FROM nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04 AS builder
+
+ENV DEPS "build-essential git ca-certificates curl unzip wget libgomp1 patchelf coreutils tcl libjemalloc-dev"
```
rafie: We should consider utilizing /opt/system-setup.py infrastructure for requirements installations, as we repeat them in CircleCI and native environments.

.circleci/config.yml
```diff
@@ -90,7 +90,24 @@ jobs:
...
+      - checkout
+      - run:
+          name: Bulding Docker
+          command: |
+            docker build -f Dockerfile.gpu-test -t redisai-gpu-test .
+      - run:
+          name: Test
+          command: |
+            mkdir -p ~/workspace/tests
+            docker run --gpus all -v $HOME/workspace/tests:/data/tests/logs -it --rm redisai-gpu-test
```
rafie: I'm not sure whether CircleCI supports volumes (-v) with docker.

.circleci/config.yml
```diff
@@ -90,7 +90,24 @@ jobs:
...
-
+  build-gpu:
+    machine:
+      docker_layer_caching: true
+      resource_class: gpu.nvidia.medium
+      image: ubuntu-1604-cuda-10.1:201909-23
+    steps:
+      - checkout
+      - run:
+          name: Bulding Docker
```
rafie: "Building"

test/basic_tests.py
hhsecond: ok

hhsecond:
@lantiga @rafie I agree with you on moving the dependency installations to `system-setup.py` but the fact that docker layer caching doesn't work on `ADD` for some reason causes all of the dependency installation to reruns on all the builds. This seems inefficient for me especially since GPU resources are costly. Let's see if CircleCI folks have some idea about this and I'll move to `system-setup` anyways

.circleci/config.yml
hhsecond: Just verified. It works on `machine` executor

hhsecond:
@lantiga @rafie 
I have removed all the dependency installations from the docker file and relying on `system-setup.py` for most of the part. But dependencies required for the Redis installations are still in the docker file. It is because docker layer caching we had enabled wasn't working as expected and putting those dependencies also to system-setup will cause the job to take quite a lot of time to finish. I have raised a complaint with them already. Let see what they have got.
Also, the GPU backend is disabled in circle CI from today and they are going to send us a contract soon.
LET'S NOT MERGE THIS TILL THE GPU SUPPORT IS BACK

lantiga:
@hhsecond Thank you Sherin, great job. Given the costs associated with building on GPU, it's really important we don't waste precious compute minutes building.

opt/build/libtorch/repack.sh
```diff
@@ -62,6 +65,7 @@ fi
 LIBTORCH_ZIP=libtorch-${PT_BUILD}-${PT_OS}-${PT_ARCH}-${PT_VERSION}.zip
 if [ ! -f $LIBTORCH_ARCHIVE ]; then
 	echo "Downloading libtorch ${PT_VERSION} ${PT_BUILD}"
+  echo $LIBTORCH_URL
```
rafie: echo can be removed.

opt/build/libtorch/repack.sh
```diff
@@ -10,13 +10,12 @@ ROOT=$HERE/../../..
...
-	GPU=no
+	GPU=0
 elif [[ "$1" == "gpu" ]]; then
-	GPU=yes
+	GPU=1
 else
-	GPU=${GPU:-no}
+	GPU=${GPU:-0}
 fi
-
```
rafie: We can keep the blank line. :-)

test/test_requirements.txt
```diff
@@ -1,3 +1,4 @@
 numpy
+scipy
```
rafie: Do we really need scipy?

get_deps.sh
```diff
@@ -169,7 +169,8 @@ if [[ $WITH_TFLITE != 0 ]]; then
 
 		LIBTFLITE_ARCHIVE=libtensorflowlite-${TFLITE_OS}-${TFLITE_ARCH}-${TFLITE_VERSION}.tar.gz
 
-		[[ ! -f $LIBTFLITE_ARCHIVE || $FORCE == 1 ]] && wget --quiet $LIBTF_URL_BASE/$LIBTFLITE_ARCHIVE
+		[[ ! -f $LIBTFLITE_ARCHIVE || $FORCE == 1 ]] && wget $LIBTF_URL_BASE/$LIBTFLITE_ARCHIVE
+		echo $LIBTF_URL_BASE/$LIBTFLITE_ARCHIVE
```
rafie: echo can be removed.

test/test_requirements.txt
lantiga: I don’t think we do

lantiga:
(@hhsecond please address Rafi’s comments before merging)

lantiga:
Hey @hhsecond, are we good to go?

PR #234, opened 2019-10-04
added noAVX cpu support for TF.1.14.0 x64 linux

june9666:
Running on a noAVX x64 cpu will crash TF.1.14.0 during load. I built and uploaded to my repo(link should be changed) libtensorflow.


june9666:
TF loads successfully, but we have "Redis version does not support SharedAPI; running without exposing C API to other modules. " issue. I think its due the lib, closing pull request, till i found a fix

PR #235, opened 2019-10-13
Add TensorFlow-Lite 2.0 backend

lantiga:
This PR adds the TFLite backend from TensorFlow 2.0

Things to complete:
- ~fix bug leading to crash when two models are set and one of them is ran, see test (this is blocking)~ DONE
- thorough GPU testing (including Edge TPU)
- arm32 build (optional)


K-Jo:
@rafie first to review intel integration, secondly ARM integration

rafie:
I'd rather follow the methodology exhibited in the ARM version of TF & PT, in which we build the library, prepare a versioned SDK archive file an upload it to S3, and then retrieve it with get_deps.sh.
The build process is done using Docker (so we have a controlled environment), and the required files harvested from the container into the SDK archive.
I'll prepare the scripts and place them into /opt/build/tflite.

PR #239, opened 2019-11-01
Selective test execution (by engine)

rafie:
If an engine is disabled on build, do not execute its tests.

### DIFF
--- a/test/basic_tests.py
+++ b/test/basic_tests.py
@@ -106,6 +106,9 @@ def test_set_tensor(env):
 
 
 def test_del_tf_model(env):
+    if not TEST_PT:
+        return
+
     test_data_path = os.path.join(os.path.dirname(__file__), 'test_data')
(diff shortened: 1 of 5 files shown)


test/basic_tests.py
```diff
@@ -144,70 +149,80 @@ def test_run_tf_model(env):
     env.assertEqual(type(exception), redis.exceptions.ResponseError)
 
     try:
-        env.execute_command('AI.MODELSET', 'm_1', 'TF',
-                            'INPUTS', 'a', 'b', 'OUTPUTS', 'mul', model_pb)
+        if TEST_TF:
```
hhsecond: Wouldn't this (and all the similar cases below) fail if `TEST_TF` is `False` since `exception` will not be raised here?

rafie: But of course. The whole test is TF so it can be controlled from the head.

PR #247, opened 2019-11-12
Patched mkdocs-material to support googletagmanager.com/gtag

rafie:


itamarhaber:
@rafi vis a vis with this, you should make a proper PR upstream to mkdocs imo.

Issue #240, opened 2019-11-05
verbose logging of "Run took %fms" is causing performance impact on the default configuration and a bad user experience on docker


filipecosta90:
If running RedisAI for the first time, or with a verbose log level ( default on RedisAI image (( non containerized redis default is notice )))
```
docker run -p 6379:6379 redisai/redisai:latest
```
we receive an enormous amount of logs, as seen on the example bellow. Considering that the information of backend inference duration will be very important in the future, we should not remove the calculation but adjust the log level in accordance or ( even better ) keep the information but dont log it ( and provide it in the future on the info command or similar )

```
1:C 05 Nov 2019 04:16:28.947 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1:C 05 Nov 2019 04:16:28.947 # Redis version=5.0.5, bits=64, commit=00000000, modified=0, pid=1, just started
1:C 05 Nov 2019 04:16:28.947 # Configuration loaded
1:M 05 Nov 2019 04:16:28.948 * Running mode=standalone, port=6379.
1:M 05 Nov 2019 04:16:28.948 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
1:M 05 Nov 2019 04:16:28.948 # Server initialized
1:M 05 Nov 2019 04:16:28.948 # WARNING you have Transparent Huge Pages (THP) support enabled in your kernel. This will create latency and memory usage issues with Redis. To fix this issue run the command 'echo never > /sys/kernel/mm/transparent_hugepage/enabled' as root, and add it to your /etc/rc.local in order to retain the setting after a reboot. Redis must be restarted after THP is disabled.
1:M 05 Nov 2019 04:16:29.077 * Module 'ai' loaded from /usr/lib/redis/modules/redisai.so
1:M 05 Nov 2019 04:16:29.077 * Ready to accept connections
2019-11-05 04:16:53.596371: I tensorflow/core/platform/cpu_feature_guard.cc:141] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.1 SSE4.2 AVX AVX2 FMA
1:M 05 Nov 2019 04:17:32.446 * <ai> RAI_ModelRun took 0.050000s
1:M 05 Nov 2019 04:17:32.447 * <ai> RAI_ModelRun took 0.000000s
[...]

Issue #92, opened 2019-03-19
Add AI.INFO command


lantiga:
We should write the equivalent of the Redis `INFO` command, to make it easy to write plugin for Prometheus, Telegraf, etc for monitoring (time to compute, number of requests, memory, state of cuda allocations, etc)

lantiga:
Use Redis module INFO API
```
RedisModule_RegisterInfoFunc
RedisModule_InfoAddSection
RedisModule_InfoBeginDictField
RedisModule_InfoEndDictField
RedisModule_InfoAddFieldString
RedisModule_InfoAddFieldCString
RedisModule_InfoAddFieldDouble
RedisModule_InfoAddFieldLongLong
RedisModule_InfoAddFieldULongLong
```

lantiga:
Decision is still implement `AI.INFO` and replicate the same info using the `INFO` API when Redis 6 is out.

lantiga:
Addressing this in #259

filipecosta90:
@lantiga closing this since it was already merged :+1:

PR #248, opened 2019-11-14
WIP: AI.INFO / INFO command required step on backend duration 

filipecosta90:
solves #240 

looking at the model info PR on redis that is already merged ( https://github.com/antirez/redis/pull/6270 ) we will still need to keep that info within some context of RedisAI which will then be used to report back in a new method with the following signature:
`
void InfoFunc(RedisModuleInfoCtx *ctx, int for_crash_report)
`
with that in mind I would suggest adding ( at least ) two fields two the `RAI_Model` struct in the same manner that redis does to have stats:
```
long long backend_calls;
long long backend_microseconds;
```
afterward, when someone called `INFO` we would report for each model the average backend_duration. basically `RAI_Model` would look like:
```
typedef struct RAI_Model {
  void* model;
  // TODO: use session pool? The ideal would be to use one session per client.
  //       If a client disconnects, we dispose of the session or reuse it for
  //       another client.
  void *session;
  RAI_Backend backend;
  char* devicestr;
  char **inputs;
  size_t ninputs;
  char **outputs;
  size_t noutputs;
  long long refCount;
  long long backend_calls;
  long long backend_microseconds;
  void* data;
} RAI_Model;
```
what do you think @lantiga ? This would be the minimal to expected metric to have. This is a simple solution now and removes the LogLine in exchange for keeping the info on the Model and reporting the average when requested.


lantiga:
Hi @filipecosta90, looks good! Since I'm into batching now, it would probably be a good thing to also report the time averaged over the batch size for that particular run.
Like, if I run a model on a batch of 5 images, and someone else runs the same model on a batch of 10 images, having an idea of the performance per sample would be informative IMHO.
Let's merge this for now and figure out the batch thing later.

filipecosta90:
> Hi @filipecosta90, looks good! Since I'm into batching now, it would probably be a good thing to also report the time averaged over the batch size for that particular run.
> Like, if I run a model on a batch of 5 images, and someone else runs the same model on a batch of 10 images, having an idea of the performance per sample would be informative IMHO.
> Let's merge this for now and figure out the batch thing later.

Hi @lantiga, @K-Jo correctly reminded us that we need to discuss whether we will integrate the information of INFO command from Redis or we will use the AI.INFO. Nonetheless, apart from where we will report it, I think that we all agree that we have that requirement. Let's take this "momentum" and bring this debate to #92.

lantiga:
Hi @filipecosta90, @K-Jo of course, let's take the discussion to #92. The rationale for going with INFO is to leverage the existing tooling and integration that look at INFO (which is probably why this new API was added to Redis).

PR #250, opened 2019-11-26
Fixed download of PyTorch 1.3.1

rafie:


get_deps.sh
```diff
@@ -16,6 +16,7 @@ if [[ $1 == --help || $1 == help ]]; then
 		Argument variables:
 		VERBOSE=1       Print commands
 		FORCE=1         Download even if present
+		WITH_DLPACK=0   Skip dlpack
```
lantiga: I don't think we can get away from DLPACK. It is not relative to the backend, it's an essential component for the RedisAI core module.

opt/build/libtorch/repack.sh
```diff
@@ -24,8 +24,7 @@ ARCH=$(python3 $ROOT/opt/readies/bin/platform --arch)
 [[ $OS == macosx ]] && export LC_ALL=en_US.UTF-8
 
 if [[ -z $PT_VERSION ]]; then
-	PT_VERSION=1.2.0
-	#PT_VERSION="latest"
+	PT_VERSION="latest"
```
lantiga: Just a confirmation: if `PT_VERSION` is set in `get_deps` (that is, if repack is called from there), then we don't set it to `latest` right?

get_deps.sh
rafie: I know that. This flag is merely for diagnostic purposes, to prevent get_deps.sh from checking it out when FORCE=1 and we're aiming for the script logic, not for the subsequent build.

opt/build/libtorch/repack.sh
rafie: That's right. We might as well bail out with an error, as we never use this script for a "floating" version.

PR #255, opened 2019-12-06
Enable AOF

lantiga:
With #157 having been merged, this PR enables AOF and runs the test suite with both RDB and AOF persistence.

TODO: activate AOF tests in CI

### DIFF
--- a/opt/Makefile
+++ b/opt/Makefile
@@ -134,7 +134,9 @@ ifneq ($(NO_LFS),1)
 endif
 	$(SHOW)set -e ;\
 	cd $(ROOT)/test ;\
+	python3 -m RLTest $(TEST_ARGS) --test basic_tests.py --module $(INSTALL_DIR)/redisai.so ;\
 	python3 -m RLTest $(TEST_ARGS) --test basic_tests.py --module $(INSTALL_DIR)/redisai.so --use-slaves ;\
+	python3 -m RLTest $(TEST_ARGS) --test basic_tests.py --module $(INSTALL_DIR)/redisai.so --use-aof ;\
 	python3 -m RLTest $(TEST_ARGS) --test double-panda.py --module $(INSTALL_DIR)/redisai.so
 
 #----------------------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


lantiga:
Odd, running tests with `useSlaves=True` passed to the Env constructor leads to the test runner to hang indefinitely after the first test passes.

@rafie @MeirShpilraien does this ring a bell?

MeirShpilraien:
@lantiga I see that you are getting the env on the test function and then create it (again) at the test itself. This actually cause 2 env's to be created at your test and then undefined behavior (when I say undefined I mean that I did not check what might happened in this case).
The way that I meant it to be use is either to get the environment at the test as argument and this way the environment will be create as requested when running the test, or to created the environment inside the test and then created it as the test need it (but not combine them both).

This way for example you can create a test that run both on single shard and with slave by just writing the test generically and then run it twice, one with the --use-slave argument and one without.

On the other hand you might want your test to run only with slave and then you need to not receive the env as an argument and just create it inside the test the way you want.

Hope it was clear enough, let me know if it helped.

lantiga:
@MeirShpilraien thank you for your answer, that clears it. When I saw the env as an argument, I thought there would be a way to augment the incoming environment with programmatically set arguments.

At this point I would rather rely on command-line arguments, so we can execute the same tests in different situations.

lantiga:
Looking for a review, then ready to merge

opt/Makefile
```diff
@@ -134,8 +134,10 @@ ifneq ($(NO_LFS),1)
 endif
 	$(SHOW)set -e ;\
 	cd $(ROOT)/test ;\
+	python3 -m RLTest $(TEST_ARGS) --test basic_tests.py --module $(INSTALL_DIR)/redisai.so ;\
 	python3 -m RLTest $(TEST_ARGS) --test basic_tests.py --module $(INSTALL_DIR)/redisai.so --use-slaves ;\
-	python3 -m RLTest $(TEST_ARGS) --test  double-panda.py --module $(INSTALL_DIR)/redisai.so
```
rafie: This was failing on macOS, so this line actually waits for the macOS CI to be enabled. If we're convinced that everything is in order there, we can remove the rest of the test artifacts. Otherwise, it may be worthwhile to just comment it out until CI is ready.

lantiga:
@K-Jo FYI I updated the pack version here

PR #259, opened 2019-12-26
AI.INFO command

lantiga:
This PR adds a new AI.INFO command to RedisAI.
Stats on each model/script key being run are accumulated and reported periodically.

TODO items:
- [x] refactor with `modelkey` as a mandatory argument (details below in the **UPDATE**)
- [x] finalize what info gets returned (added `ERRORS`, evaluating `HISTOGRAM`)
- [x] report batch size per call
- [x] add info reset command (`AI.INFO modelkey RESET`) to clean up all stat keys and start fresh
- [x] write tests
- [x] write docs

**UPDATE** (credit @K-Jo): avoid getting stats from all models with a single command (response potentially unbounded, it looks like a scan), but have users call `AI.INFO` with a `modelkey` as the argument. This way a) the stats will have the same lifetime as the respective models (~we can collocate stats directly in the model struct and get rid of the module-level dict~ UPDATE: we will still keep the stats in the dict to avoid accessing keyspace in the callback); b) responses will be bounded and specific for each model. Users will also be in charge of calling `RESETSTAT` on the model stats.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -2,8 +2,6 @@
 
 ## AI.CONFIG LOADBACKEND
 
-### AI.CONFIG LOADBACKEND Example
-
 Load a DL/ML backend.
 
 By default, RedisAI starts with the ability to set and get tensor data, but setting and running models and scritps requires a computing backend to be loaded. This command allows to dynamically load a backend by specifying the backend identifier and the path to the backend library. Currently, once loaded, a backend cannot be unloaded, and there can be at most one backend per identifier loaded.
@@ -15,6 +13,8 @@ AI.CONFIG LOADBACKEND <backend_identifier> <location_of_backend_library>
 * allowed backend identifiers are: TF (TensorFlow), TORCH (PyTorch), ONNX (ONNXRuntime).
 
 It is possible to specify backends at the command-line when starting `redis-server`, see example below.
+
+### AI.CONFIG LOADBACKEND Example
 
 > Load the TORCH backend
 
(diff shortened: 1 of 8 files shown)


filipecosta90:
@lantiga @K-Jo @gkorland  OSS look and feel. I think that @itamarhaber can be of extreme help here 
```
https://redis.io/commands/config-resetstat
https://redis.io/commands/info
```

src/redisai.c
```diff
@@ -721,6 +724,17 @@ struct RedisAI_RunInfo {
   RAI_Error* err;
 };
 
+struct RedisAI_RunStats {
+  RedisModuleString *key;
+  int type; // model or script
+  RAI_Backend backend;
+  char* devicestr;
+  long long duration_us;
```
filipecosta90: @lantiga based on total duration in us and total calls we can calculate the average model/script run time. My question goes if the average is enough right now or do we want extended metrics like quantiles?

src/redisai.c
```diff
@@ -721,6 +724,17 @@ struct RedisAI_RunInfo {
   RAI_Error* err;
 };
 
+struct RedisAI_RunStats {
```
filipecosta90: keeping track of the error counts that ocurred during backend running might be of use. what do you think @lantiga ?
please see https://github.com/tensorflow/serving/issues/462#issuecomment-367594255

src/redisai.c
```diff
@@ -1332,6 +1398,55 @@ int RedisAI_ScriptSet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
+    RedisModule_ReplyWithSimpleString(ctx, "SAMPLES");
+    if (rstats->type == 0) {
+      RedisModule_ReplyWithLongLong(ctx, rstats->samples);
+    }
+    else {
+      RedisModule_ReplyWithLongLong(ctx, -1);
+    }
+    RedisModule_ReplyWithSimpleString(ctx, "CALLS");
+    RedisModule_ReplyWithLongLong(ctx, rstats->calls);
+
```
filipecosta90: missing calls that resulted in error. see the above comment. 
IMHO also missing the overall model running latency histogram which is something easy to implement.

src/redisai.c
lantiga: Quantiles would certainly be interesting, @filipecosta90 do you have a suggestion for computing running quantiles?

lantiga: (see e.g. https://sites.cs.ucsb.edu/~suri/psdir/ency.pdf)

src/redisai.c
lantiga: Yes they would, noted

src/redisai.c
filipecosta90: @lantiga we're working on the usage of histograms specifically target for latency measurements on other projects ( HDR histogram ). I will revise your link and we can discuss this afterwards on a meeting :) Independently of the code used, the whole idea of extended metrics is a must IMHO.

src/redisai.c
lantiga: Thank you @filipecosta90, do you have a reference for an efficient implementation?

lantiga: Like: https://github.com/HdrHistogram/HdrHistogram_c ?

src/redisai.c
lantiga: Done

lantiga:
Added tests and docs, the PR is ready to land, except for the histograms.
@filipecosta90 please let me know what you'd like to do regarding that (see above in the comments). Thanks!

src/redisai.c
filipecosta90: hi there @lantiga, yes. Just tested locally the time on adding a value to the histogram and it should be around ~5ns - so we can be very confident that this histogram is extremely efficient. See:
https://github.com/HdrHistogram/HdrHistogram_c/pull/76#issue-366594040

docs/commands.md
```diff
@@ -234,3 +234,42 @@ If needed, input tensors are copied to the device specified in `AI.SCRIPTSET` be
...
+```sql
+AI.INFO <model_or_script_key> RESETSTAT
+```
+
+The command can be called on a key until that key is removed using `MODELDEL` or `SCRIPTDEL`.
+
+### AI.INFO Example
+
+```sql
+AI.INFO amodel
```
filipecosta90: @lantiga can we include the example output here?

lantiga: Done

PR #271, opened 2020-02-17
[wip] enabling coverage of CPU code

filipecosta90:
@rafie very preliminary but it enables running the tests within Cmake and then coverage it.


filipecosta90:
closing in favor of new cleaned version

PR #272, opened 2020-02-17
Usereplicas.fix

filipecosta90:
quick fix on tests to enforce waiting on replica before reads ( required to make sure we're correctly using RLTest --use-replicas )


filipecosta90:
it has mixed commits. splitting

PR #296, opened 2020-02-20
[add] tests refactoring per backend

filipecosta90:
@lantiga and @rafie this splits the tests per backend and the common ones. It also extends coverage and ensures slave sync when required. The whole purpose is to clean and make the tests clear and remove code duplicates 
-  test/includes.py - has the methods to load models, check the env variables, etc... 
-  test/tests_common.py  - tests for common commands like tensorset 
-  test/tests_onnx.py - tests for ONNX backend
-  test/tests_pytorch.py  - tests for torch backend
-  test/tests_tensorflow.py  - tests for TF backend
-  test/tests_tflite.py  - tests for tflite backend

leak fixes:
- fixes #280 
- fixes #285

PR #297, opened 2020-02-20
Test.refactor.coverage

filipecosta90:
to be discussed after #296

### DIFF
--- a/opt/system-setup.py
+++ b/opt/system-setup.py
@@ -22,7 +22,7 @@ class RedisAISetup(paella.Setup):
 
         if self.os == 'linux':
             self.install("ca-certificates")
-        self.install("git cmake unzip wget patchelf awscli")
+        self.install("git cmake unzip wget patchelf awscli valgrind")
         self.install("coreutils") # for realpath
 
     def debian_compat(self):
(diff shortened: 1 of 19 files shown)


opt/system-setup.py
```diff
@@ -22,7 +22,7 @@ def common_first(self):
 
         if self.os == 'linux':
             self.install("ca-certificates")
-        self.install("git cmake unzip wget patchelf awscli")
+        self.install("git cmake unzip wget patchelf awscli valgrind")
```
rafie: This will fail on macOS. Better to install Valgrind in:
```
def linux(self):
    self.install("valgrind")
```

rafie:
This can be way off the current implementation, but as we don't use CMake to run tests, I prefer to drive coverage data collection and report generation outside CMake (the build part is of course ok but it's relatively modest in configuration). TimeSeries has it implemented this way.

filipecosta90:
> This can be way off the current implementation, but as we don't use CMake to run tests, I prefer to drive coverage data collection and report generation outside CMake (the build part is of course ok but it's relatively modest in configuration). TimeSeries has it implemented this way.

Agree @rafie. will check TS and change accordingly :)

filipecosta90:
closing in favor of #299

PR #305, opened 2020-03-12
Bump ONNXRuntime to 1.2.0 (CUDA requirement 10.1)

lantiga:


lantiga:
ONNXRuntime backend not loading on GPU due to
```
482:M 14 Mar 2020 00:11:24.753 # <ai> Could not load ONNX backend from /build/bin/linux-x64-release/install-gpu/backends/redisai_onnxruntime/redisai_onnxruntime.so: libcublas.so.10: cannot open shared object file: No such file or directory
```

Issue #306, opened 2020-03-12
Add ability to TAG (or assign LABELS to) a model


lantiga:
We need a way to verify the identity of a running model (beyond the key itself).

We need to decide whether we need full key-value information (like `LABELS label value` as Redis TimeSeries), or just a string TAG.

One nice side-effect is that now we can aggregate information at the AI.INFO level by also taking the TAG into consideration (stats from a model deployed at one key where a previous model was set wouldn't mix in the interval). The TAG would be reported as part of the AI.INFO stats.

Issue #304, opened 2020-03-11
Meta argument for MODELGET


hhsecond:
Similar to the META argument for tensors, it's really helpful to have an API that returns the meta-information of a model such as backend, device, etc.

gkorland:
@hhsecond that should be part of the `AI.INFO <key>`

Issue #303, opened 2020-03-11
List all the models deployed


hhsecond:
Currently, we don't have any mechanism to figure out the models deployed using RedisAI. While it is important to have an API to list all the models, It might be a good idea to have an API that lists scripts and tensors as well

lantiga:
We will implement a command to list models and scripts, along with their TAGS (see #306) for identification. In order to do that, we can leverage on the shard-level dictionary that we maintain for AI.INFO.
I don't think we should list tensors - they could come in a very large number, it would be equivalent to a keyscan.

PR #307, opened 2020-03-14
Add TAG to models and revamp MODELGET

lantiga:
Addresses:
* #306 by adding an optional `TAG <tag>` argument to `AI.MODELSET` and `AI.SCRIPTSET`; tags are returned by `AI.MODELGET` and `AI.SCRIPTGET` and they are reported as part of `AI.INFO`
* #304 by adding a `META` or `BLOB` argument to `AI.MODELGET` and `META` or `SOURCE` argument to `AI.SCRIPTGET` to get information like backend, device, tag, with or without the actual model blob or script source
* #303 by adding two new commands `AI.MODELLIST` and `AI.SCRIPTLIST`, that return key and tag for each model or script respectively

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -165,23 +165,23 @@ AI.MODELRUN resnet18 INPUTS image12 OUTPUTS label12
     The execution of models will generate intermediate tensors that are not allocated by the Redis allocator, but by whatever allocator is used in the backends (which may act on main memory or GPU memory, depending on the device), thus not being limited by maxmemory settings on Redis.
 ---
 
-## AI.MODELLIST
+## AI._MODELLIST
 
-NOTE: `MODELLIST` is EXPERIMENTAL and might be removed in future versions.
+NOTE: `_MODELLIST` is EXPERIMENTAL and might be removed in future versions.
 
 List all models. Returns a list of (key, tag) pairs.
 
 ```sql
-AI.MODELLIST
+AI._MODELLIST
 ```
 
-### MODELLIST Example
+### _MODELLIST Example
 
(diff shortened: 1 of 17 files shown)


lantiga:
@K-Jo @gkorland @filipecosta90 @hhsecond 
Since we're still unsure about the `MODELLIST` and `SCRIPTLIST` commands, can we still merge and mark them `experimental` or remove them from the API and merge the rest?

hhsecond:
It'd help me to make the MLFlow integration easier if we have MODELLIST. I think we should keep it and mark it as experimental

lantiga:
I agree with this, let's get moving. @hhsecond can you take a look at the PR?

filipecosta90:
@lantiga and @hhsecond i believe that we should split this PR to exclude #303 . Lets approve #306 and #304 changes and discuss further the listing/scanning :)

lantiga:
@filipecosta90 how long do you think it will take? I'd rather not have scan hold up MLFlow integration. We can always switch listing to scan at a later time.

lantiga:
We could eventually avoid having those commands in the documentation, so that MLFlow integration can use them but they are not advertised. And maybe prepend them with `_`, like `AI._MODELLIST` to further reinforce the fact you shouldn't use them.

filipecosta90:
>  prepend them with `_`, like `AI._MODELLIST` to further reinforce the fact you shouldn't use them.

agree :+1: :) 
Until we figure out the new API it's ok if he'll push forward with these experimental APIs

hhsecond:
Awesome, thanks guys

PR #308, opened 2020-03-15
[add] added performance section describing aibench

filipecosta90:
this PR adds a performance section describing aibench. As soon as we publish the RE results we can make reference of it.
The goal of this page is to provide guidance of how to measure, what to collect,etc.


K-Jo:
@filipecosta90 can we add the diagrams ?

PR #310, opened 2020-03-25
Batching crashtest

hhsecond:
Test case for testing the batching crash

### DIFF
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -98,7 +98,8 @@ queueItem *queueEvict(queue *queue, queueItem *item) {
     queue->back->next = NULL;
   }
   else {
-    item->prev->next = item->next->prev;
+    item->prev->next = item->next;
+    item->next->prev = item->prev;
   }
 
   item->next = NULL;
--- a/test/tests_tensorflow.py
+++ b/test/tests_tensorflow.py
@@ -1,4 +1,5 @@
 import redis
+from functools import wraps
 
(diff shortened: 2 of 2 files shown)


lantiga:
Thank you @hhsecond !

lantiga:
I'm wondering why the non-gpu build is not crashing. Can you confirm this is the case?

hhsecond:
Confirming that It's crashing on CPU. Tests probably misbehaving because of the timings. I have added a thread join for syncing the execution flow (Although both GPU and CPU tests are failing already, right?)

hhsecond:
Ok, that's just weird. All the tests are passing now

hhsecond:
Ok, so if you take to look at the logs, the server is still crashing. But the test conditions are met because it tries to fetch the result from a key that was filled from a previous operation. I have changed the test. It should fail now. But something still keeps me thinking is, how'd the crashed server auto-healed for the rest of the test cases in the pipeline. Would you have any idea @lantiga ?

lantiga:
@hhsecond I think I got it: `queueEvict` had a big bug in it.
Great job with the repro BTW, it would have been really hard to spot the bug without it.

Note: I had to port the test to multiprocessing so I can now kill the pending `MODELRUN` process. Otherwise the test passes but we get an exception afterwards, because `redis-server` is terminated by the runner but that request is still pending.

PR #313, opened 2020-03-31
tests for model serialization for pytorch, tensorflow lite and onnx

filipecosta90:
@lantiga this PR adds tests for model serialization for all pytorch, tensorflow lite and onnx backends. However, on tensorflow and pytorch some assertions are not true and I wanted to double-check with you if this is not a bug.
For each model saving test we MODELSET, then MODELGET, then SAVE, then MODELGET. We have 3 serialized versions of the model that should all match:
- model loaded from disk
- model retrieved from MODELGET from memory
- model retrieved from MODELGET after an RDB load


lantiga:
Hi @filipecosta90, the thing is that we re-serialize models at each RDB save or MODELGET, each time serializing it from whatever memory they are in (CPU or GPU). BTW: is an expensive operation, especially from GPU, so the plan is to keep a cache of the serialized model around for saves (see #122), at the cost of more memory consumption. Or maybe we should only do it if we load the model on GPU (note that for ONNXRuntime we already keep a cache because it has no serialization API).

Anyhow, the re-serialization is not guaranteed to produce the same model, so I'm expecting that the serialized model is not the same as the one we load. Of course it would be if we cached.

filipecosta90:
@lantiga after what we've discussed changed the tests on torch and tf to compare only models metadata and to ensure that the model runs after loading from rdb. It seems that TF has an issue

filipecosta90:
@lantiga removed the failing test for tensorflow. If CI passes we can merge this one after review. Let's tackle the tensorflow issue on another PR/issue.

PR #315, opened 2020-04-02
Add Keras model test with TF2 model

lantiga:
TF2 supports freezing models, which can ben ran on TF1.15.

This PR adds a test on a MNIST-like model generated with Keras, exported through TF2 and that is ran on RedisAI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,8 @@ RedisAI currently supports PyTorch (libtorch), Tensorflow (libtensorflow), Tenso
 | 0.9.0   | 1.3.1   | 1.14.0     | 2.0.0  | 1.0.0         |
 | master  | 1.4.0   | 1.15.0     | 2.0.0  | 1.0.0         |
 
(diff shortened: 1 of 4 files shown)


lantiga:
@hhsecond you might want to add this use case to your export library

hhsecond:
Awesome. I'll do that. Assuming they have added it in 2.1?
Also, I believe this doesn't mean we can run all 2.x models in 1.15 because there are 2.x specific ops.

I know this because  [we have built](    if path.is_dir():  # Expecting TF SavedModel) the same functionality using 1.15 earlier. i.e a SavedModel directory from 2.x can be opened using 1.15 and then convert it to frozen model from there. And I had issues converting few models from 2.x to 1.15

hhsecond:
Maybe with this new utility, they might convert it to a backward-compatible representation? Very unlikely

lantiga:
This is slightly different though since it's TF2 doing the freezing. Since the frozen model is only useful for running in TF1, I'm expecting TF2 to fail if there's no corresponding op in GraphDef, so at least it would fail early on the TF2 side.
Can we try challenging the freezing?

hhsecond:
That's a fair point. Lemme find the models I tried with the old technique.

PR #320, opened 2020-04-08
Extended testing for tensorset and modelrun

filipecosta90:
@lantiga just some extended testing for tensorset and modelrun. On tensorset it includes a multiprocess test with tensorset multiple times also in the same key ( to force eviction ).
It brings more confidence to merge this one and then #317.


filipecosta90:
@lantiga tests passing, ready to merge :)

PR #321, opened 2020-04-09
Code refactoring, consistent error messages, and RedisAI_ModelRun_RedisCommand optimizations

filipecosta90:
Current WIP
- [add] simplified argument processing on RedisAI_ModelRun_RedisCommand (only 1 full-loop vs 5)
- [add] code refactoring to reduce duplicates and reduce complexity. 
- [add] enforced error reply string comparison on tests. 
- [fix] error messages consistent across commands for tensor getting, etc...
- removes usage of `RedisModule_AutoMemory(ctx)` from `RedisAI_TensorGet_RedisCommand`, `RedisAI_ModelSet_RedisCommand`, `RedisAI_ModelGet_RedisCommand`, `RedisAI_ModelDel_RedisCommand`, `RedisAI_ModelRun_RedisCommand`, `RedisAI_ScriptDel_RedisCommand`
- moves `queue` from `redisai.c` to `util/queue.c`

### DIFF
--- a/src/util/queue.h
+++ b/src/util/queue.h
@@ -1,0 +1,41 @@
+
+#include <pthread.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "../redisai_memory.h"
+#include "redismodule.h"
+
+#ifndef __QUEUE_H
+#define __QUEUE_H
+
+typedef struct queueItem {
+  struct queueItem *next;
+  struct queueItem *prev;
+  void *value;
+} queueItem;
+
+typedef struct queue {
+  queueItem *front;
+  queueItem *back;
+  void (*free)(void *ptr);
+  unsigned long len;
+} queue;
+
+queue *queueCreate(void);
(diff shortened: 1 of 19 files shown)


src/util/queue.h
```diff
@@ -0,0 +1,41 @@
...
+queue *queueCreate(void);
+void queuePush(queue *queue, void *value);
+queueItem *queuePop(queue *queue);
+queueItem *queueFront(queue *queue);
+queueItem *queueNext(queueItem *item);
+queueItem *queueEvict(queue *queue, queueItem *item);
+long long queueLength(queue *queue);
+void queueRelease(queue *queue);
+
+typedef struct RunQueueInfo {
```
lantiga, line 34: I think `RunQueueInfo` doesn't belong to `util`

lantiga: We can keep it in redisai.c for the time being, but later on we can factor it out in a runqueue.c (not in util) along with code related to aynchronous processing.

Issue #203, opened 2019-09-05
RedisAI needs to manage the number of background threads for backends


hhsecond:
Libtorch uses all the available cors (including hyperthreaded ones) which causes performance degradation.

K-Jo:
@hhsecond, can we make an abstraction layer in general between the backends and redisai to specify which resources can be used by the redis process?

hhsecond:
We could. Currently, the only libtorch has this problem though

filipecosta90:
@lantiga @K-Jo @hhsecond and @gkorland I believe this is of most importance, specifically for solutions with multiple shards on the same machine. 
I would stick to the same nomenclature of TF and Pytorch and add two loadtime configurations:
- INTRA_OP_PARALLELISM ( multiple threads within the ops )
- INTER_OP_PARALLELISM ( explicitly use a global (per process) thread pool )
If set to 0 or not set the system will not force and use the default per backend configs. 
At least [PyTorch](https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html#runtime-api) and [TF](https://github.com/tensorflow/tensorflow/issues/13853#issuecomment-466408718)

filipecosta90:
#322 adds the possibility to specify both configurations(INTRA_OP_PARALLELISM and INTER_OP_PARALLELISM). Only enforced for now on Tensorflow backend.

filipecosta90:
@hhsecond @gkorland @lantiga and @K-Jo I've done a quick test with the added configurations of interop and intra-op parallelism of #322, and even on a local dev environment this plays a big role in the achievable ops/sec and overall inference latency. 
For OSS RedisAI with financialnet aibench we see that keeping the background threads limited is beneficial. This is directly bound to the model complexity ( which enforces the need to be able to set/controll the degree of parallelism on the backends ). 

INTER_OP_PARALLELISM | INTRA_OP_PARALLELISM | infnerences/sec | RTT q50 latency (ms)
-- | -- | -- | --
1 | 1 | 16867 | 0.46
2 | 1 | 11666 | 0.65
2 | 2 | 11657 | 0.66
4 | 2 | 10506 | 0.72

hhsecond:
Ideally, doing this for PyTorch is straight-forward with these two functions

```
torch::set_num_threads
torch::set_num_interop_threads
```
[...]

Issue #88, opened 2019-03-12
Add `DAGRUN` (or `PIPERUN`) command


lantiga:
A common pattern is enqueuing multiple `SCRIPTRUN` and `MODELRUN` commands. Setting input tensors, storing intermediates of `*RUN` commands and storing outputs all in keyspace means that all written keys will be AOF'd and replicated even if they will be disposed of shortly after.

## DAGRUN

A general solution to this problem is to have a `DAGRUN` (or `PIPERUN`, we'll see about the name - it is a DAG rather than a pipe, so it should probably be named correctly) command that allows to run a sequence of `AI.*` commands, e.g.:

```sql
AI.DAGRUN TENSORSET ~a~ FLOAT 2 VALUES 2 3 => \
          TENSORSET ~b~ FLOAT 2 VALUES 2 3 => \
          MODELRUN foo INPUTS ~a~ ~b~ OUTPUTS ~c~ => \
          SCRIPTRUN bar baz INPUTS ~c~ OUTPUTS d
```

Note the key names between `~ ~`: these are _volatile_ keys (`~a~` is volatile because it has small wings :-) ). A volatile key is not set into keyspace, but it just lives in-memory for the duration of the command, after which is deallocated. In the command above, `~a~`, `~b~`, `~c~` are volatile, so they don't touch keyspace (and are not replicated). Only the output key `d` is stored in keyspace, for later retrieval.

## Relationship to MODELRUN that returns results

This design supersedes #85, which proposed a `MODELRUN` variant that returns the output. This can be now obtained by:
```sql
AI.DAGRUN MODELRUN foo INPUTS a b OUTPUTS ~c~ => TENSORGET ~c~
```
or, without touching keyspace at all
```sql
AI.DAGRUN TENSORSET ~a~ FLOAT 2 VALUES 2 3 => \
          TENSORSET ~b~ FLOAT 2 VALUES 2 3 => \
          MODELRUN foo INPUTS ~a~ ~b~ OUTPUTS ~c~ => \
          TENSORGET ~c~
```

## Advantages

Apart from the obvious convenience, there are several advantages with this design:
- once a `DAGRUN` command is sent, we can apply DAG optimization strategies; for instance, in case of ensambles, we can execute different `MODELRUN` subcalls in parallel (this was not possible before because each call was blocking on the client) and then join on the results to execute a further `SCRIPTRUN` that computes the ensembled outputs;
[...]

PR #322, opened 2020-04-12
WIP - AI.DAGRUN, commands code refactoring and improvement

filipecosta90:
kickoff version that enables #88 

- [x] refactored ensureRunQueue to pass also RunQueueInfo
- [x] refactored TENSORGET, TENSORSET to be re-used by AI.DAGRUN
- [x] refactored MODELRUN's `RedisAI_ModelRun_RedisCommand` ( part 1 of 3 (main thread portion )) to be capable of use local tensors as input
- [X] add local tensors structure to `RedisAI_RunInfo` 
- [X] refactored MODELRUN's RedisAI_Run_Reply ( part 3 of 3 (main thread portion )) to be capable of use local tensors as output (requires changes of `RedisAI_RunInfo` first )
- [X] do the persist on unblock callback
- [X] enable running more than one blocking command inside dag reusing the same runinfo data 
- [X] simple tensorset |> modelrun |> tensorget with and without touching the keyspace via writes ( only read via LOAD )
- [X] discover the DAGRUN device queue from the arguments of MODELRUN. If no MODELRUN default to CPU
- [X] Move the modelctx from rinfo to DagOp. Instead of a single mctx and sctx pointer, it holds an array of (mctx | sctx) pointers (within RAI_DagOp). To be gradually adopted on modelrun and scriptrun ( for now only on dagrun )
- [X] Write after read test on DAGs local context
- [X] proper negative testing
- [X] **leak check on negative testing**
- [X] rebase on current master (minor batching refactoring)
- [X] **leak check on positive testing**
- [X] **removed complexity and potential unsafe behaviour of tensorset**
Extra (enable further optimizations):
- [X] enable INTRA_OP_PARALLELISM and INTER_OP_PARALLELISM load time configurations. Only enforced for now on TF backend.   ( kickoff for #203 )


src/redisai.c
```diff
@@ -1125,129 +1212,154 @@ int RedisAI_ModelList_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
+ * The request is queued and evaded asynchronously from a separate thread. The
+ * client blocks until the computation finishes.
+ *
+ * 1. clone inputs as needed in the main thread (only the alternative is to
+ * lock)
+ * 2. spawn the new thread for running the model
+ * 3. have reply callback put the data back into the key
+ * 
+ * This way we avoid any race condition. The only gotcha is making sure no one
+ * overwrites the model until it's done computing.
```
lantiga, line 1326: I find this comment a bit misleading:
`RAI_ModelRunCtxCreate` will shallow copy the model, so since this command runs in the main thread, there's no need for locking or waiting because run context will hold the model that was found in the key at the time this portion of the command was executed. If a new model is set while the run is waiting in the queue, the model is disposed at the key level, but the old model is retained in the run context and runs. It is then deallocated when run context is deallocated.

filipecosta90: @lantiga agree. This was an old comment that just moved from inside the function to the function description. 
https://github.com/RedisAI/RedisAI/blob/master/src/redisai.c#L1139
Should we PR this change specifically on a different pull request, merge it on master and then update here? Or just updating here is enough?

lantiga: Right, let’s just update here. I’d like to fast track this work anyway

lantiga:
Looking very good so far.

I propose to refactor RunInfo, so that instead of a single mctx and sctx pointer, it holds an array of (mctx | sctx) pointers. I think it's better to do it now than wait too long and have to change code pervasively again. This will open the door to the full DAG implementation which at this point I'd like to get done shortly. I'll be happy to do that as soon as you're done with the rest, just let me know.

We need to be careful with the semantics of values we reply with from the local context. Say we have:
```
TENSORSET foo ...
MODELRUN m1 INPUTS foo OUTPUTS bar
TENSORGET bar
MODELRUN m2 INPUTS foo OUTPUTS bar
```
then we should get `bar` after `m1` has run, not `m2`.

Ok, this uses two MODELRUN's, but in the current code it's the same thing as
```
TENSORSET foo ...
TENSORSET bar ...
TENSORGET bar ...
MODELRUN m INPUTS foo OUTPUTS bar
TENSORGET bar ...
```
I would expect `bar` to have two different values in output. What do you think?

lantiga:
Another comment: should we call the command simply `AI.RUN`?

filipecosta90:
> I propose to refactor RunInfo
Agree @lantiga . I'm also refactoring the `RedisAI_DagRun_RedisCommand` to only parse the `LOAD` and `PERSIST` and split the commands and pass all logic to the background thread as you've mentioned

src/redisai.c
```diff
@@ -1826,104 +900,18 @@ int RedisAI_Config_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv, i
...
-  if (number_loaded_keys != n_keys) {
-    RedisModule_WrongArity(ctx);
-    return -1;
-  }
-  return argpos;
-}
-
 /**
- * AI.DAGRUN [LOAD <nkeys> key1 key2... ] [PERSIST <nkeys> key1 key2... ] |>
+ * AI.DAGRUN device [LOAD <nkeys> key1 key2... ] [PERSIST <nkeys> key1 key2... ] |>
```
lantiga, line 904: I don't think we should have a DAG-wise device. This limits DAG in the future.
If we limit the current implementation to an individual device (I imagine thats' because this way everything is in the same thread), let's detect it and thow a "non supported yet" error.

filipecosta90: @lantiga we need a device to add to the queue. I believe we're bound to this

test/tests_dag.py
```diff
@@ -0,0 +1,367 @@
...
+        env.assertEqual(ret, 0 )
+
+        # assert that classificationTensor does not exist
+        ret = con.execute_command("EXISTS classificationTensor:{}".format(
+            tensor_number))
+        env.assertEqual(ret, 0 )
+        tensor_number = tensor_number + 1
+
+
+def test_dag_modelrun_financialNet_no_writes_multiple_modelruns(env):
```
filipecosta90, line 315: @lantiga and @gkorland , multiple modelruns within the same dag working as expected

src/redisai.c
filipecosta90: > Looking very good so far.
> 
> I propose to refactor RunInfo, so that instead of a single mctx and sctx pointer, it holds an array of (mctx | sctx) pointers. I think it's better to do it now than wait too long and have to change code pervasively again. This will open the door to the full DAG implementation which at this point I'd like to get done shortly. 
> 

@lantiga the latest commit makes usage of this change for DAGRUN, so that instead of a single mctx and sctx pointer, it holds an array of (mctx | sctx) pointers (within RAI_DagOp). To be gradually adopted on modelrun and scriptrun ( for now and to be able to fully test and harden it only on dagrun ). 
What do you think?

src/redisai.c
filipecosta90: As you've mentioned @lantiga bouding ourselfs to parsing the device string would block further use-cases. For now doing as we've discussed and getting the device string from the model info via parsing. If multiple devices are present on the same DAG reply with error for now

test/tests_dag.py
```diff
@@ -0,0 +1,368 @@
...
+        "AI.TENSORGET tensor1 VALUES"
+
+    ret = con.execute_command(command)
+    env.assertEqual(ret, [b'OK', [b'FLOAT', [1, 2], [b'5', b'10']]])
+
+    ret = con.execute_command("AI.TENSORGET tensor1 VALUES")
+    env.assertEqual(ret, [b'FLOAT', [1, 2], [b'5', b'10']])
+
+
+def test_dag_local_multiple_tensorset_on_same_tensor(env):
```
filipecosta90, line 103: > We need to be careful with the semantics of values we reply with from the local context. Say we have:
> 
> ```
> TENSORSET foo ...
> MODELRUN m1 INPUTS foo OUTPUTS bar
> TENSORGET bar
> MODELRUN m2 INPUTS foo OUTPUTS bar
> ```
> 
> then we should get `bar` after `m1` has run, not `m2`.
> 
> Ok, this uses two MODELRUN's, but in the current code it's the same thing as
> 
> ```
> TENSORSET foo ...
> TENSORSET bar ...
> TENSORGET bar ...
> MODELRUN m INPUTS foo OUTPUTS bar
> TENSORGET bar ...
> ```
> 
> I would expect `bar` to have two different values in output. What do you think?

@lantiga agree and we have it working with the current implementation. Please revise the `test_dag_local_multiple_tensorset_on_same_tensor` added test

src/redisai.c
filipecosta90: solved in ab37bbe7320c8b588639afd24645825d06e89f4d

lantiga:
LGTM at a first complete review pass. I'd say let's do a last review pass after rebasing on master + leak fixes.

src/background_workers.c
```diff
@@ -0,0 +1,197 @@
...
+        if (current_batchsize == 0 || current_batchsize >= batchsize) {
+          break;
+        }
+
+        queueItem *next_item = item->next;
+
+        while (next_item != NULL) {
+          RedisAI_RunInfo *next_rinfo = (RedisAI_RunInfo *)next_item->value;
+
+          if (RAI_RunInfoBatchable(rinfo, next_rinfo) == 0) {
```
lantiga, line 138: Here we need to prevent a DAG run info to be considered batchable. I'm pushing a commit to make that happen.

lantiga: BTW, we should note this limitation and extend batching to DAG as well at some point. This will require some DAG refactoring: individual MODELRUN commands should be placed in the queue, so they have a chance to be batched. We'll have to make batching logic slightly more complicated to avoid the order of DAG commands to change for an individual client due to batching.
We can leave this out for now, but I'll open an issue.

Issue #329, opened 2020-04-18
Revamp GET methods


lantiga:
Based on a discussion between @itamarhaber and myself, we should restructure the `*GET` method API:

The main issue is not having a command that just returns the raw BLOB (or SOURCE in SCRIPTGET) data; this is limiting for working with the CLI, because we can't just redirect the output to a file

The solution is to have [ALL | META | BLOB] as output options, where BLOB only returns the raw data, while ALL returns META + BLOB. This way we can make the behavior of *GET methods consistent.

Since we're at it, we should also return arrays of *bulk strings* and not simple strings in these replies.

> I suggest the following: AI.*GET will always return an array of bulk strings. These are alternating field-value pairs. Each will have the ability to accept [ALL | META | BLOB] with ALL being the default.

lantiga:
Actually, instead of [ALL | META | BLOB], since TENSOR has VALUES and it wouldn't be clear what needs to be returned with ALL, we could settle for

```
TENSORGET key [META] [BLOB | VALUES]
```
so to get just meta of `foo` one goes
```
TENSORGET key META
```
to get just the data buffer
```
TENSORGET key BLOB
```
to get both
```
TENSORGET key META BLOB
```

Same thing for MODELGET
```
MODELGET key [META] [BLOB]
```
and
```
SCRIPTGET key [META] [SOURCE]
```

PR #332, opened 2020-04-19
Revamp GET methods (breaking change for clients)

lantiga:
Addresses #329 

Specifically, with this PR, get methods have a uniform signature and behavior
```
TENSORGET key [META] [BLOB | VALUES]
MODELGET key [META] [BLOB]
SCRIPTGET key [META] [SOURCE]
```

If META is not specified, all methods return a raw binary BLOB, an array VALUES or a raw SOURCE string. This helps, for instance, with dumping a buffer out of a model or a tensor from the CLI.

If only META is specified, all methods return meta data in [key, val, key, val] form. This is a breaking change with respect to the previous behavior (TENSORGET would return [dtype, shape, blob] and not ['dtype', dtype, 'shape', shape, 'blob', blob]), so **THIS CHANGE WILL BREAK CLIENTS AND USER CODE**.

However, this is our last call to make the API uniform and ready for RESP3.

Other notable changes:
- keys in [key, value, ...] responses are now lowercase (including `AI.INFO`)
- all commands returning strings other than "OK" or "ERR" now return bulk strings 
- the experimental `_MODELLIST` command has been renamed to `_MODELSCAN` (to prepare for the upcoming `TENSORLIST` data type)

Tests and docs have been updated.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -41,20 +41,76 @@ AI.TENSORSET foo FLOAT 2 2 VALUES 1 2 3 4
 Get a tensor.
 
 ```sql
-AI.TENSORGET tensor_key [BLOB | VALUES | META]
+AI.TENSORGET tensor_key [META] [BLOB | VALUES]
 ```
 
 * tensor_key - Key for the tensor
+* META - Return tensor meta data (data type and shape)
 * BLOB - Return tensor content as a binary buffer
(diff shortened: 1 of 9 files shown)


lantiga:
Genius here forgot to merge origin/master prior to creating patch. Re-pushing soon.

lantiga:
Ok, rebased on master

PR #338, opened 2020-04-23
Add possibility to provide a model in chunks

lantiga:
While testing on of the HuggingFace models (1.3 GB), `MODELSET` was failing due to the fact that Redis strings are limited to 512 MB.

This PR gives the possibility to provide a large model in chunks, overcoming the limitation (at the cost of a transient allocation + memcpy of a large chunk of memory, but for now there are no alternatives):
```
    model = f.read()
chunk_size = 500 * 1024 * 1024
model_chunks = [model[i:i + chunk_size] for i in range(0, len(model), chunk_size)]
ret = con.execute_command('AI.MODELSET', 'm', 'TORCH', DEVICE, 'BLOB', *model_chunks)
```

Note that the API has been changed. One now has to specify `BLOB` to signal the start of chunks:
```
AI.MODELSET foo TORCH CPU BLOB ...
```

For consistency, the same was done for SCRIPTSET, with the addition of `SOURCE`:
```
AI.SCRIPTSET foo CPU SOURCE ...
```

@hhsecond: we need to update the Python client both with `BLOB` and `SOURCE`, as well as chunking (detect if blob is larger than 512 MB and chunk it tansparently)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Note that Redis config is located at `/usr/local/etc/redis/redis.conf` which can
 
 On the client, set the model
 ```sh
-redis-cli -x AI.MODELSET foo TF CPU INPUTS a b OUTPUTS c < test/test_data/graph.pb
+redis-cli -x AI.MODELSET foo TF CPU INPUTS a b OUTPUTS c BLOB < test/test_data/graph.pb
 ```
 
 Then create the input tensors, run the computation graph and get the output tensor (see `load_model.sh`). Note the signatures:
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -123,7 +123,7 @@ AI.TENSORGET foo META VALUES
 Set a model.
 
 ```sql
-AI.MODELSET model_key backend device [TAG tag] [BATCHSIZE n [MINBATCHSIZE m]] [INPUTS name1 name2 ... OUTPUTS name1 name2 ...] model_blob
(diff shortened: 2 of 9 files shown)


lantiga:
Thank you @itamarhaber.
What do you think about adding a FILE option, in addition to BLOB, where the file is loaded by Redis in a location relative to some configurable PATH?
This would have the advantage that RDB and AOF would just reference that one, without growing 1.3GB in size upon serialization.
It would make sense for really large models.

cc @filipecosta90

filipecosta90:
> What do you think about adding a FILE option, in addition to BLOB, where the file is loaded by Redis in a location relative to some configurable PATH?
> This would have the advantage that RDB and AOF would just reference that one, without growing 1.3GB in size upon serialization.
> It would make sense for really large models.
> 
> cc @filipecosta90

@lantiga imho that would break the entire idea around portability of the DB files, and open the door for more complex errors that I believe should not be a responsibility of the module. The module should use the persistency options provided by redis and allow as single point of entrance for data the commandHandler/commands. What do you think?

lantiga:
@filipecosta90 that has always been my idea too. But these models are huge, so I was thinking about possible ways to anticipate RDB/AOF serialization issues.
I'm going to merge this now and we can think about it in the future (and the API now allows for it without further backward breakages).
I suspect this will come out as a request some of the use cases.

filipecosta90:
> @filipecosta90 that has always been my idea too. But these models are huge, so I was thinking about possible ways to anticipate RDB/AOF serialization issues.
> I'm going to merge this now and we can think about it in the future (and the API now allows for it without further backward breakages).
> I suspect this will come out as a request some of the use cases.

agree =)

PR #347, opened 2020-05-01
PT 1.5 & TF 1.15.0 w/cuda10.1 + ORT 1.2.0

rafie:


itamarhaber:
@lantiga README and docs should be updated to include the info as well

PR #349, opened 2020-05-02
packages

gkorland:


docs/quickstart.md
```diff
@@ -21,8 +21,9 @@ docker run -p 6379:6379 --gpus all -it --rm redisai/redisai:latest-gpu
 You can compile and build the module from its source code. The [Developer](developer.md) page has more information about the design and implementation of the RedisAI module and how to contribute.
 
 ### Prerequisites
-* CUDA needs to be installed for GPU support.
+* Packages: git, python3, make, wget, g++, & unzip 
 * CMake 3.0 or higher needs to be installed.
+* CUDA needs to be installed for GPU support.
```
rafie, line 26: We currently use Docker for GPU=1 builds (utilizing Nvidia SDK image). I'm not sure we can recommend installing CUDA SDK with a straight face. That is, our build do work if CUDA is installed on the system (like in AWS Deep Learning instances), but I've never tried installing CUDA on a clean, non-Nvidia-recommended machine.

PR #353, opened 2020-05-04
Redis 6 in builds

rafie:


lantiga:
Happy to merge if we are good

itamarhaber:
BTW, we should continue also testing on 5.x imo (latest is 5.0.9).

lantiga:
> BTW, we should continue also testing on 5.x imo (latest is 5.0.9).

+1 (also see my previous comment)

PR #356, opened 2020-05-05
The display title of the site has a typo

evelynmitchell:
https://oss.redislabs.com/redisai/ shows 
RedisAI - A Server for Maching and Deep Learning Models
s/Maching/Machine/ is the fix.
I added a whitespace at the end of this file, for the fix, but it is not here, it's elsewhere.


lantiga:
Thank you @evelynmitchell! Funny it's been under our nose and we didn't realize it.

PR #368, opened 2020-05-08
[WIP] add C API reference to docs using auto tool 

filipecosta90:
image

to update the C api reference:

```
cd docs
python3 generate_llapi_reference.py 
Running doxygen step...
Running doxybook step...
Searching on ./../src/redisai.c for REGISTER_API macro...
Associating 44 registered llapi functions with their code comments...
Saving markdown to api_reference.md...
```
to test it, at the root project folder:

```
mkdocs serve
```


filipecosta90:
revised with @itamarhaber , we need to build the llapi .md file on every commit instead. 
Doing the changes before merging.

filipecosta90:
@itamarhaber I've moved to generate the C API reference on the fly as you've recommended. 

There is still another thing to discuss which is the position of the C API reference on the Nav. From what I recall I believe both @DvirDukhan and @K-Jo mentioned it would be better fitted into Developer. From talking with @gkorland yesterday I believe he has also the same position. 
I see pros (and cons) on choosing either of the nav positions - so I believe it's a matter of consensus =)
WDYT @lantiga @itamarhaber @hhsecond ?

lantiga:
@filipecosta90 I agree with having the C API under Developer

lantiga:
Approved, pending the Dockercloud issue (I didn't understand what is failing - I get a 404 from the issue)

filipecosta90:
> Note: Dockerhub builds fail due to known issue when adding submodules. It resolves itself when merged to master.

@lantiga I believe the docker build issue is related to the update from master ( which updates the dependencies ) and follows what @rafie stated in #377 . WDYT @rafie ?

PR #384, opened 2020-05-25
Fixes AI.MODELGET to return INPUTS,OUTPUTS, BATCHSIZE, and MINBATCHSIZE

### LINKED ISSUE #378: AI.MODELGET needs to return INPUTS,OUTPUTS, BATCHSIZE, and MINBATCHSIZE
We need the above properties to properly allow for:
- a full copy of models via modelget->modelset

filipecosta90:
fixes #378 and updates docs accordingly. This should not break any client (if it does then we need to adjust the clients not to be so restrictive for future changes)


filipecosta90:
> LGTM

@lantiga already merged master on it. Can you check it after the gpu build passes?

PR #387, opened 2020-05-25
Enable chunking for large models in AOF, RDB, replication and MODELGET

lantiga:
Addresses #386 

Tests are missing, we need a strategy for downloading a large (512 MB) model from CI.

Note: this PR changes the format of RDB. Versioning is handled, ie older RDB are still loaded correctly.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -244,44 +244,31 @@ int RAI_configParamParse(RedisModuleCtx *ctx, const char *key,
   else if (strcasecmp((key), "THREADS_PER_QUEUE") == 0) {
     ret = RedisAI_Config_QueueThreads(rsval);
     if (ret == REDISMODULE_OK) {
-      char *buffer = RedisModule_Alloc(
-          (3 + strlen(REDISAI_INFOMSG_THREADS_PER_QUEUE) + strlen((val))) *
-          sizeof(*buffer));
-      sprintf(buffer, "%s: %s", REDISAI_INFOMSG_THREADS_PER_QUEUE, (val));
-      RedisModule_Log(ctx, "notice", buffer);
-      RedisModule_Free(buffer);
+      RedisModule_Log(ctx, "notice", "%s: %s",
+                      REDISAI_INFOMSG_THREADS_PER_QUEUE,
+                      (val));
     }
   } else if (strcasecmp((key), "INTRA_OP_PARALLELISM") == 0) {
     ret = RedisAI_Config_IntraOperationParallelism(rsval);
     if (ret == REDISMODULE_OK) {
-      char *buffer = RedisModule_Alloc(
-          (3 + strlen(REDISAI_INFOMSG_INTRA_OP_PARALLELISM) + strlen((val))) *
-          sizeof(*buffer));
-      sprintf(buffer, "%s: %lld", REDISAI_INFOMSG_INTRA_OP_PARALLELISM,
-              getBackendsIntraOpParallelism());
-      RedisModule_Log(ctx, "notice", buffer);
-      RedisModule_Free(buffer);
+      RedisModule_Log(ctx, "notice", "%s: %lld",
+                      REDISAI_INFOMSG_INTRA_OP_PARALLELISM,
+                      getBackendsIntraOpParallelism());
     }
   } else if (strcasecmp((key), "INTER_OP_PARALLELISM") == 0) {
     ret = RedisAI_Config_InterOperationParallelism(rsval);
     if (ret == REDISMODULE_OK) {
-      char *buffer = RedisModule_Alloc(
-          (3 + strlen(REDISAI_INFOMSG_INTER_OP_PARALLELISM) + strlen((val))) *
-          sizeof(*buffer));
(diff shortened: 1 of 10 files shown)


src/redisai.c
```diff
@@ -456,7 +456,12 @@ int RedisAI_ModelGet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
 
   if (meta && blob) {
     RedisModule_ReplyWithCString(ctx, "blob");
-    RedisModule_ReplyWithStringBuffer(ctx, buffer, len);
+    const size_t n_chunks = len / RAI_CHUNK_LEN + 1;
+    RedisModule_ReplyWithArray(ctx, (long)n_chunks);
```
filipecosta90, line 460: @lantiga this will break the clients correct? IMO we also need the change documented as soon as we merge this.

lantiga: Yes it will. Part of this review was also to coordinate on this aspect.

lantiga: Done

src/config.c
```diff
@@ -228,6 +273,15 @@ int RAI_configParamParse(RedisModuleCtx *ctx, const char *key,
...
       RedisModule_Free(buffer);
     }
+  } else if (strcasecmp((key), "MODEL_CHUNK_SIZE") == 0) {
+    RedisAI_Config_ModelChunkSize(rsval);
+    char *buffer = RedisModule_Alloc(
+        (3 + strlen(REDISAI_INFOMSG_MODEL_CHUNK_SIZE) + strlen((val))) *
+        sizeof(*buffer));
+    sprintf(buffer, "%s: %lld", REDISAI_INFOMSG_MODEL_CHUNK_SIZE,
+            getModelChunkSize());
+    RedisModule_Log(ctx, "notice", buffer);
```
DvirDukhan, line 283: `void REDISMODULE_API_FUNC(RedisModule_Log)(RedisModuleCtx *ctx, const char *level, const char *fmt, ...)`
I think you can do
`RedisModule_Log(ctx, "notice", "%s: %lld", REDISAI_INFOMSG_MODEL_CHUNK_SIZE,
            getModelChunkSize());`

src/config.c
```diff
@@ -228,6 +273,15 @@ int RAI_configParamParse(RedisModuleCtx *ctx, const char *key,
       RedisModule_Log(ctx, "notice", buffer);
       RedisModule_Free(buffer);
     }
+  } else if (strcasecmp((key), "MODEL_CHUNK_SIZE") == 0) {
+    RedisAI_Config_ModelChunkSize(rsval);
```
DvirDukhan, line 277: Check error code?

src/redisai.c
```diff
@@ -456,7 +467,18 @@ int RedisAI_ModelGet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
...
+    if (n_chunks > 1) {
+      RedisModule_ReplyWithArray(ctx, (long)n_chunks);
+      for (size_t i=0; i<n_chunks; i++) {
+        size_t chunk_len = i < n_chunks - 1 ? chunk_size : len % chunk_size;
+        RedisModule_ReplyWithStringBuffer(ctx, buffer + i * chunk_size, chunk_len);
+      }
+    }
+    else {
+      RedisModule_ReplyWithStringBuffer(ctx, buffer, len);
+    }
```
DvirDukhan, line 481: Reorder code or migrate to a function?

src/version.h
```diff
@@ -6,4 +6,9 @@
 /* API versions. */
 #define REDISAI_LLAPI_VERSION 1
 
+/* This is the major / minor encver, to be used as
+   argument to RM_CreateDataType, which expects
+   encver to be < 1024 */
+static const long long RAI_ENC_VER_MM = RAI_ENC_VER == 999999 ? 1023 : RAI_ENC_VER / 100;
```
DvirDukhan, line 12: As far as I understand, current (master) RAI_ENC_VER is 0? 
How do you plan to maintain updates to RAI_ENC_VER in the objects RDB decoders?

lantiga: Current master is `RAI_ENC_VER` `999999`, while for 1.0 `RAI_ENC_VER` is `10000` (major `1`, minor `00`, patch `00`).

src/config.c
lantiga: Done

src/config.c
lantiga: Done

src/redisai.c
lantiga: Done

src/version.h
DvirDukhan: ok
So there is a need to update a minor version once this update is merged, right?
For now, the decoding of objects from RDB is checking if RAI_ENC_VER > 100 for the new partitioned decode, otherwise it will decode as a single buffer. This is ok for now since there are only two "official" versions. Once there will be additional modifications in encoding/decoding objects to RDB I suggest moving to a dedicated decoding function per encoding version.

lantiga: Yes, if we release this will need to be a minor (1.1.0)

Issue #392, opened 2020-05-27
Support variadic INPUTS in SCRIPTRUN (and potentially MODELRUN)


lantiga:
Syntax could be

```
SCRIPTRUN foo bar INPUTS [ a b ] c OUTPUTS d
```

which translates to a function signature
```
List[Tensor] Tensor -> Tensor
```
in TorchScript

PR #395, opened 2020-06-01
Add support for variadic arguments to SCRIPT

lantiga:
This PR addresses #392.
A slightly different command API was chosen to keep things efficient.

We use the `$` sign to signal that `INPUTS` following that sign will be passed to the script as a list:
```
redis-cli -x AI.SCRIPTSET foo CPU SOURCE < foo.py
redis-cli AI.TENSORSET a FLOAT 1 1 VALUES 1
redis-cli AI.TENSORSET b FLOAT 1 1 VALUES 1
redis-cli AI.TENSORSET c FLOAT 1 1 VALUES 1
redis-cli AI.SCRIPTRUN foo foo INPUTS a $ b c OUTPUTS d
redis-cli AI.TENSORGET d META VALUES
```

The corresponding script will need to be type-annotated with `List[Tensor]` for the last argument
```
def foo(a : Tensor, args: List[Tensor]):
    return a + torch.stack(args).sum()
```

Within the script, `args` will act as a standard Python list containing tensors.

Test has been added and docs have been updated.

### DIFF
--- a/test/tests_pytorch.py
+++ b/test/tests_pytorch.py
@@ -583,6 +583,66 @@ def test_pytorch_scriptrun_errors(env):
         env.assertEqual(type(exception), redis.exceptions.ResponseError)
 
 
+def test_pytorch_scriptrun_errors(env):
+    if not TEST_PT:
+        env.debugPrint("skipping {} since TEST_PT=0".format(sys._getframe().f_code.co_name), force=True)
+        return
+
+    con = env.getConnection()
+
+    test_data_path = os.path.join(os.path.dirname(__file__), 'test_data')
+    script_filename = os.path.join(test_data_path, 'script.txt')
+
+    with open(script_filename, 'rb') as f:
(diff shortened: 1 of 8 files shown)


test/tests_pytorch.py
```diff
@@ -426,6 +426,61 @@ def test_pytorch_scriptrun(env):
...
+
+    ensureSlaveSynced(con, env)
+
+    if env.useSlaves:
+        con2 = env.getSlaveConnection()
+        values2 = con2.execute_command('AI.TENSORGET', 'c', 'VALUES')
+        env.assertEqual(values2, values)
+
+
 def test_pytorch_scriptrun_errors(env):
```
filipecosta90, line 484: @lantiga can we add some negative tests on variadic scriptrun?

lantiga: Done!

PR #397, opened 2020-06-03
added low level api return redis types

DvirDukhan:
This PR allows external modules to get RedisAI key types, so they can read AI keys from keyspace or write AI low-level API objects to keyspace, at their will.
This also makes #394 redundant.

@lantiga please tell me if there is a need to update low-level API version


lantiga:
Looks good, but I don't understand why CI is failing. I'll try to build on my end

PR #398, opened 2020-06-03
Multi-platform build

rafie:


.circleci/config.yml
```diff
@@ -19,69 +30,86 @@ commands:
           command: |
             ./opt/readies/bin/getpy3
             BREW_NO_UPDATE=1 ./opt/system-setup.py
-            # git clone git://github.com/antirez/redis.git --branch 6.0.1; (cd redis; make malloc=libc -j $(nproc); make install); redis-server --version
+            # ./opt/readies/bin/getredis -v 6 --force
```
rafie, line 33: make sure Redis is built with `malloc=libc`.

filipecosta90:
LGTM @rafie :clap: 
we just need to change the required steps names on GH correct? Who has permissions to do it (@lantiga @gkorland )?

.circleci/config.yml
lantiga: Hi @rafie, looks great. I'm pretty sure `libc` is the default on macOS

lantiga:
@gkorland I've been told you're the one doing this for other modules. Do you have time to take care of this?

PR #400, opened 2020-06-04
Llapi updates

DvirDukhan:
This PR allows:
1. Retrieve model, script, and tensor Redis types from LLAPI so other modules can read/write AI keys 
2. Added LLAPI for variadic script input - modified the existing input and variadic input for integrity verification before adding new input. Added failure test for double variadic input.

### DIFF
--- a/src/script.c
+++ b/src/script.c
@@ -335,3 +335,7 @@ void RedisAI_ReplyOrSetError(RedisModuleCtx *ctx, RAI_Error *error, RAI_ErrorCod
         RedisModule_ReplyWithError(ctx, errorMessage);
     }
 }
+
+RedisModuleType *RAI_ScriptRedisType(void) {
+    return RedisAI_ScriptType;
+}
--- a/src/script.c
+++ b/src/script.c
@@ -164,8 +164,27 @@ static int Script_RunCtxAddParam(RAI_ScriptRunCtx* sctx,
   return 1;
 }
 
-int RAI_ScriptRunCtxAddInput(RAI_ScriptRunCtx* sctx, RAI_Tensor* inputTensor) {
-  return Script_RunCtxAddParam(sctx, sctx->inputs, inputTensor);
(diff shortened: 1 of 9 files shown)


src/script.h
```diff
@@ -67,9 +67,25 @@ RAI_ScriptRunCtx* RAI_ScriptRunCtxCreate(RAI_Script* script,
  *
  * @param sctx input RAI_ScriptRunCtx to add the input tensor
  * @param inputTensor input tensor structure
+ * @param err error data structure to store error message in the case of
+ * failures
  * @return returns 1 on success ( always returns success )
```
filipecosta90, line 72: we should update the @return statement

src/script.h
```diff
@@ -67,9 +67,25 @@ RAI_ScriptRunCtx* RAI_ScriptRunCtxCreate(RAI_Script* script,
...
+ * For each Allocates a RAI_ScriptCtxParam data structure, and enforces a shallow copy of
+ * the provided input tensor, adding it to the input tensors array of the
+ * RAI_ScriptRunCtx.
+ *
+ * @param sctx input RAI_ScriptRunCtx to add the input tensor
+ * @param inputTensors input tensors array
+ * @param len input tensors array len
+ * @param err error data structure to store error message in the case of
+ * failures
+ * @return returns 1 on success ( always returns success )
```
filipecosta90, line 86: same as above

src/script.c
```diff
@@ -309,10 +333,7 @@ int RedisAI_Parse_ScriptRun_RedisCommand(RedisModuleCtx *ctx,
                         return -1;
                     }
                 }
-                if (!RAI_ScriptRunCtxAddInput(*sctx, inputTensor)) {
-                    RedisAI_ReplyOrSetError(ctx, error, RAI_ESCRIPTRUN, "ERR Input key not found");
-                    return -1;
-                }
+                if (!RAI_ScriptRunCtxAddInput(*sctx, inputTensor, error)) return -1;
```
lantiga, line 336: Why did we remove the input key error here?

lantiga:
@filipecosta90 I'm ok to merge, but in perspective (sooner than later) we should have a C++ test suite

PR #406, opened 2020-06-08
Avoid splitting outputs in batches when nbatches == 1

### LINKED ISSUE #401: MODELRUN results missing output values
I'm using this ONNX YOLOv3 model including the pre- and post processing functions:
https://github.com/onnx/models/tree/master/vision/object_detection_segmentation/yolov3

If this this model is used with RedisAI it consistently returns only one detection. When I configure the same model using seperate ONNX Runtime instance, multiple detections are returned. I also  tried using YOLO TensorFlow model with RedisAI, but the results are still invalid. I also tried with different input images, with no luck.

From the outputs below I can see that RedisAI returns only the first detection. Is the problem a RedisAI bug or my is my implementation faulty?

I'm running latest redisai/redisai docker image. Both ONNX Runtime instances are version 1.2.0

AI.INFO returns 0 errors.

### Test image
dog

### Preprocessing code

```
import onnxruntime
import numpy as np
from PIL import Image
from redisai import Client as RaiClient
import redisai

def letterbox_image(image, size):
    iw, ih = image.size
    w, h = size
    scale = min(w/iw, h/ih)
    nw = int(iw*scale)
    nh = int(ih*scale)

    image = image.resize((nw,nh), Image.BICUBIC)
    new_image = Image.new('RGB', size, (128,128,128))
    new_image.paste(image, ((w-nw)//2, (h-nh)//2))
    return new_image

def preprocess(numpy_img):
    img = Image.fromarray(numpy_img)
    model_image_size = (416, 416)
    boxed_image = letterbox_image(img, tuple(reversed(model_image_size)))
    image_data = np.array(boxed_image, dtype='float32')
    image_data /= 255.
    image_data = np.transpose(image_data, [2, 0, 1])
    image_data = np.expand_dims(image_data, 0)
    return image_data

```

### RedisAI implementation
```
yolo_model = "./yolov3-10.onnx"

with open(yolo_model, 'rb') as f:
    model = f.read()

con = redisai.Client(host='localhost', port=6380)
con.modelset('model', "ONNX", 'cpu', model)

pil_image = Image.open("./dog.jpg")
numpy_img = np.array(pil_image)
numpy_img = preprocess(numpy_img)

side_size = 416
image_size = np.array([side_size, side_size], dtype=np.float32) \
    .reshape(1, 2)

con.tensorset(
    'input_1',
    numpy_img,
    dtype="float32",
    shape=(1, 3, side_size, side_size)
    )

con.tensorset(
    'image_shape',
    image_size,
    dtype="float32",
    shape=(1, 2)
    )

con.modelrun(
    'model',
    inputs=["input_1", "image_shape"],
    outputs=['boxes', 'scores', 'classes']
    )

#classes should contain 3 detections
classes = con.tensorget('classes')
print(classes)
```
### RedisAI outputs
`[[  0   1 253]]`

### ONNX Runtime
```
yolo_model = ("./yolov3-10.onnx")
sess = onnxruntime.InferenceSession(yolo_model)

pil_image = Image.open("./dog.jpg")
numpy_img = np.array(pil_image)
numpy_img = preprocess(numpy_img)

side_size = 416
image_size = np.array([side_size, side_size], dtype=np.float32) \
    .reshape(1, 2)
result = sess.run(
    [],
    {
        "input_1": numpy_img,
        "image_shape": image_size
    }
    ) 

classes = result[2]
print(classes)
```

### ONNX Runtime outputs
`[[   0    1  253]
 [   0    7 1111]
 [   0   16  322]]`

lantiga:
Fixes #401 

TORCH and TFLITE didn't have the issue, while ONNX and TF did.

It might also be a good idea to add a test targeting this use case (not with full Yolo to avoid including a 200+MB model, only with a tiny model simulating a 1-batch input and 3-batch output).

### DIFF
--- a/src/backends/torch.c
+++ b/src/backends/torch.c
@@ -93,9 +93,9 @@ int RAI_ModelRunTorch(RAI_ModelRunCtx** mctxs, RAI_Error *error) {
 
   size_t batch_sizes[nbatches];
   size_t batch_offsets[nbatches];
+  size_t total_batch_size = 0;
 
   if (nbatches > 1) {
-    size_t total_batch_size = 0;
     if (array_len(mctxs[0]->inputs) > 0) {
       for (size_t b=0; b<nbatches; ++b) {
         batch_sizes[b] = RAI_TensorDim(mctxs[b]->inputs[0].tensor, 0);
@@ -147,7 +147,7 @@ int RAI_ModelRunTorch(RAI_ModelRunCtx** mctxs, RAI_Error *error) {
     }
     RAI_Tensor* output_tensor = RAI_TensorCreateFromDLTensor(outputs_dl[i]);
     if (nbatches > 1) {
-      if (outputs_dl[i]->dl_tensor.shape[0] != nbatches) {
+      if (outputs_dl[i]->dl_tensor.shape[0] != total_batch_size) {
         RAI_SetError(error, RAI_EMODELRUN, "ERR Model did not generate the expected batch size");
         return 1;
       }
--- a/src/backends/torch.c
+++ b/src/backends/torch.c
@@ -147,6 +147,10 @@ int RAI_ModelRunTorch(RAI_ModelRunCtx** mctxs, RAI_Error *error) {
     }
     RAI_Tensor* output_tensor = RAI_TensorCreateFromDLTensor(outputs_dl[i]);
     if (nbatches > 1) {
+      if (outputs_dl[i]->dl_tensor.shape[0] != nbatches) {
+        RAI_SetError(error, RAI_EMODELRUN, "ERR Model did not generate the expected batch size");
+        return 1;
+      }
       for (size_t b=0; b<nbatches; b++) {
         mctxs[b]->outputs[i].tensor = RAI_TensorCreateBySlicingTensor(output_tensor, batch_offsets[b], batch_sizes[b]);
       }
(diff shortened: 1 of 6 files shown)


src/backends/torch.c
```diff
@@ -147,6 +147,10 @@ int RAI_ModelRunTorch(RAI_ModelRunCtx** mctxs, RAI_Error *error) {
     }
     RAI_Tensor* output_tensor = RAI_TensorCreateFromDLTensor(outputs_dl[i]);
     if (nbatches > 1) {
+      if (outputs_dl[i]->dl_tensor.shape[0] != total_batch_size) {
+        RAI_SetError(error, RAI_EMODELRUN, "ERR Model did not generate the expected batch size");
```
filipecosta90, line 151: @lantiga should add the test to it ( it would pass coverage in that manner I believe )

filipecosta90: solved on latest commit

lantiga:
@filipecosta90 I don't understand the failing codecov/patch, what's the issue there?

filipecosta90:
> @filipecosta90 I don't understand the failing codecov/patch, what's the issue there?
> 
> Probably it's the batch size check with TF and ONNX in [7ad0d18](https://github.com/RedisAI/RedisAI/commit/7ad0d18e5a5c2481225c0e1b8eac0e43e9462404)

I believe the reason it fails is due to patch coverage ( added code is 65% ) even tough on overall the project coverage will increase. IMO we can merge and address this in the future via unit testing ( same as other corner cases that are difficult to test on flows )

Issue #370, opened 2020-05-10
Add support for multiple devices from within an individual DAG


lantiga:

filipecosta90:
Fixed by #424

Issue #328, opened 2020-04-18
Extend batching to DAG commands


lantiga:
When #322 gets merged, DAG won't support batching. That is, a `MODELRUN` within a `DAGRUN` will not have the opportunity to be batched with other `MODELRUN` commands within the DAG or within the overall queue.

Enabling batching DAG commands will require some refactoring (not major). Right now (in #322) a DAG command corresponds to a single entry in the queue. This setting is simpler, but the batching logic doesn't see the individual `MODELRUN` commands within the DAG.

To extend batching to DAG `MODELRUN` commands, we will need to break down the entire DAG in individual commands and place them in the queue. This way we'll be able to batch them.
The batching logic will need to evolve slightly, to take care that commands in the same DAG don't change the order of their execution due to batching.

That is because in normal (non-DAG) operation, since `MODELRUN` is blocking, batching will not affect the ordering of commands seen by the individual client. However, with DAG, we need to ensure that `MODELRUN |> TENSORSET |> MODELRUN` doesn't become `MODELRUN |> MODELRUN |> TENSORSET` as an effect of batching.

lantiga:
Closed by #482

PR #424, opened 2020-06-28
Multi-device parallel dag

lantiga:
This PR addresses #370, it brings
- the ability to define DAGs involving multiple devices for SCRIPTRUN and MODELRUN
- parallel execution of commands within a DAG if they are assigned to different devices

The way DAGs are handled has been refactored. This opens the door for #328 (enabling batching for individual DAG commands).

~This is a WIP, a few items are still TODO:~
- [x] the implementation so far doesn't handle reusing (overwriting) key names, they need to be unique; we need to add local key mangling to solve this transparently
- [x] analyze performance and make sure threads are spin-locking correctly
- [x] make sure there are no crashes
- [x] increase testing
- [x] analyze potential leaks
- [x] documentation

The way we allow DAG operations to run on different devices in parallel (when possible) is the following: instead of running the whole DAG in one swoop as in the previous implementation, the DAG run info is created on one queue/device and shallow copied (appropriately) across other queues/devices as indicated by the DAG specification. A DAG mutex is shared across all copies.
The DAG run info is placed on the queue for each device and evicted for execution. Execution happens one DAG op at a time: once the individual op has executed, it is marked as such and the DAG run info is placed back on the queue.
The current un-executed op is checked for its inputs. If all inputs are found in the tensor context, then the DAG op can be executed. If not, the execution quits and control is given back to the worker. If there are other items in the queue the op is placed after the next item.
When all ops for a device have been executed, the DAG is not placed back on the queue. When all ops in a DAG have been executed or an error occurs, the client is unblocked.


src/dag.c
```diff
@@ -495,48 +661,207 @@ int RedisAI_DagRunSyntaxParser(RedisModuleCtx *ctx, RedisModuleString **argv,
...
 
-  pthread_mutex_lock(&run_queue_info->run_queue_mutex);
-  queuePush(run_queue_info->run_queue, rinfo);
-  pthread_cond_signal(&run_queue_info->queue_condition_var);
-  pthread_mutex_unlock(&run_queue_info->run_queue_mutex);
+  array_free(devices);
+  array_free(rinfo_copies);
 
   return REDISMODULE_OK;
 }
```
DvirDukhan, line 867: add a line

src/dag.c
```diff
@@ -495,48 +661,207 @@ int RedisAI_DagRunSyntaxParser(RedisModuleCtx *ctx, RedisModuleString **argv,
...
+    AI_dictEntry *entry = AI_dictNext(iter);
+    while (entry) {
+      char *key = (char *)AI_dictGetKey(entry);
+      AI_dictEntry *mangled_entry = AI_dictFind(mangled_tensors, key);
+      if (!mangled_entry) {
+        return RedisModule_ReplyWithError(ctx,
+                                          "ERR PERSIST key cannot be found in DAG");
+      } 
+      int *instance = AI_dictGetVal(mangled_entry);
+      RedisModuleString *mangled_key = RedisModule_CreateStringPrintf(ctx, "%s%04d", key, *instance);
```
DvirDukhan, line 781: possible leak?

src/dag.c
```diff
@@ -495,48 +661,207 @@ int RedisAI_DagRunSyntaxParser(RedisModuleCtx *ctx, RedisModuleString **argv,
...
   }
-  // If the queue does not exist, initialize it
-  if (ensureRunQueue(deviceStr, &run_queue_info) == REDISMODULE_ERR) {
-    RAI_FreeRunInfo(ctx, rinfo);
-    return RedisModule_ReplyWithError(ctx,
-                                      "ERR Queue not initialized for device");
+  
+  const char* master_device = rinfo->dagOps[array_len(rinfo->dagOps)-1]->devicestr;
+
+  RedisAI_RunInfo **rinfo_copies = array_new(RedisAI_RunInfo*, 10);
```
DvirDukhan, line 829: allocate with capacity as the number of devices

src/dag.c
```diff
@@ -198,28 +368,29 @@ int RedisAI_DagRun_Reply(RedisModuleCtx *ctx, RedisModuleString **argv,
...
-                RAI_SafeAddDataPoint(rstats,currentOp->duration_us,1,0,0);
-                RedisModule_ReplyWithSimpleString(ctx, "OK");
-            }
-            break;
+      case REDISAI_DAG_CMD_SCRIPTRUN: {
+        rinfo->dagReplyLength++;
+        struct RedisAI_RunStats *rstats = NULL;
+        const char *runkey = RedisModule_StringPtrLen(currentOp->runkey, NULL);
+        RAI_GetRunStats(runkey,&rstats);
+        if (currentOp->result == REDISMODULE_ERR) {
```
filipecosta90, line 384: Add testcase for it. coverage report pointing this line as not covered

src/background_workers.c
```diff
@@ -181,22 +206,54 @@ void *RedisAI_Run_ThreadMain(void *arg) {
...
+          pthread_mutex_lock(evicted_rinfo->dagMutex);
+          dagError = *evicted_rinfo->dagError;
+          pthread_mutex_unlock(evicted_rinfo->dagMutex);
+        }
+        if (use_local_context == 1 && dag_complete == 0 && !dagError) {
+          if (dag_progress) {
+            queueUnpop(run_queue_info->run_queue, evicted_rinfo);
+          }
+          else {
+            if (queueLength(run_queue_info->run_queue) > 0) {
```
filipecosta90, line 237: testcase for it. codecov report showcasing we're not covering this

src/background_workers.c
```diff
@@ -181,22 +206,54 @@ void *RedisAI_Run_ThreadMain(void *arg) {
...
+      for (long long i = 0; i < array_len(evicted_items); i++) {
+        RedisAI_RunInfo *evicted_rinfo = (RedisAI_RunInfo *)(evicted_items[i]->value);
+        const int use_local_context = evicted_rinfo->use_local_context;
+        int dagError = 0;
+        if (use_local_context == 1) {
+          pthread_mutex_lock(evicted_rinfo->dagMutex);
+          dagError = *evicted_rinfo->dagError;
+          pthread_mutex_unlock(evicted_rinfo->dagMutex);
+        }
+        if (use_local_context == 1 && dag_complete == 0 && !dagError) {
```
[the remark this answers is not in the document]
lantiga: The thing is that `dagError` is a pointer that is shared between worker threads (there are copies of rinfo in every pertinent queue that share some of the items, like `dagMutex`, `dagError` and the `dagOps` - see https://github.com/RedisAI/RedisAI/blob/multidevice_dag/src/run_info.c#L147), so we indeed need to acquire the lock. The rest of the condition needs to happen only if no errors were generated in the step.
Would you simplify it further

src/background_workers.c
```diff
@@ -181,22 +206,54 @@ void *RedisAI_Run_ThreadMain(void *arg) {
...
+              queueItem *next_item = queuePop(run_queue_info->run_queue);
+              RedisAI_RunInfo *next_rinfo = (RedisAI_RunInfo *)next_item->value;
+              queueUnpop(run_queue_info->run_queue, evicted_rinfo);
+              queueUnpop(run_queue_info->run_queue, next_rinfo);
+            }
+            else {
+              queueUnpop(run_queue_info->run_queue, evicted_rinfo);
+              // Sleep for a millisecond since there's nothing else
+              // on the queue that has a chance to run
+              usleep(1000);
```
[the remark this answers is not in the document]
lantiga: This is the case where there's nothing in the queue but the item that just returned without completing, due to the fact that inputs were not computed yet. I'm sleeping just to give a chance to other worker threads to compute those inputs.
Do you think we should go without and just have one thread spin without sleeping?

src/dag.c
lantiga: Nice catch, now that we mangle we need to free the `inkeys` and `outkeys` strings when freeing the ops.

src/dag.c
lantiga: Sorry I'm not sure I get this one

lantiga: Oh I get it now :-)

src/script.c
```diff
@@ -167,26 +167,16 @@ static int Script_RunCtxAddParam(RAI_ScriptRunCtx* sctx,
 }
 
 int RAI_ScriptRunCtxAddInput(RAI_ScriptRunCtx* sctx, RAI_Tensor* inputTensor, RAI_Error* err) {
-    if(sctx->variadic != -1) {
-        RAI_SetError(err, RAI_EBACKENDNOTLOADED, "ERR Already encountered a variable size list of tensors");
-        return 0;
-    }
-    return Script_RunCtxAddParam(sctx, &sctx->inputs, inputTensor);
+  return Script_RunCtxAddParam(sctx, &sctx->inputs, inputTensor);
```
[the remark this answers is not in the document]
lantiga: Same as above

lantiga:
During coverage increase I unearthed a race condition that can lead to a crash.

There's a portion of a test in `test/tests_dag.py`, function `test_dagrun_modelrun_multidevice_resnet_ensemble_alias`, that has been commented out.

What happens is the following:

```
model1 -> script (with error)
model2 ---> 
```

The script errors out and unblocks, but model2 is still running. We should indeed only unblock when all threads have done processing. This already happens under normal circumstances, but when errors are generated in-between with certain DAG topologies this is not explicitly checked for. We need to take care of this prior to merging the PR.

lantiga:
The edge case has now been fixed. Unblocking is now controlled through ref counting.

src/script.c
```diff
@@ -167,26 +167,16 @@ static int Script_RunCtxAddParam(RAI_ScriptRunCtx* sctx,
...
-    for(size_t i=0; i < len; i++){
-        Script_RunCtxAddParam(sctx, &sctx->inputs, inputTensors[i]);
-    }
-    return 1;
+  int res;
+  for (size_t i=0; i < len; i++) {
+    res = Script_RunCtxAddParam(sctx, &sctx->inputs, inputTensors[i]);
+    if (res != 1) return res;
+  }
+  return 1;
```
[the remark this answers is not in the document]
lantiga: I fixed this by making AddInputList behave as before, but allowing one to also accrue variadic arguments using AddInput. The alternative was quite convoluted

Issue #434, opened 2020-07-21
Define `getkeys-api` on AI.DAGRUN


gkorland:
Define `getkeys-api` on AI.DAGRUN

PR #435, opened 2020-07-21
Add getkeys-api to the DAGRUN command

lantiga:
Addresses #434


gkorland:
@lantiga I don't see where you call `RedisModule_KeyAtPos` and `RedisModule_IsKeysPositionRequest`

lantiga:
You are right, my bad.
After further consideration though, we will need to defer this to after merging #424: in the new DAGRUN PR we factor out parsing all kets before blocking, which makes it a lot easier to implement the missing functions.

lantiga:
In scope: #160 #161

filipecosta90:
closing this one in favor of #438 , which already addresses this issue for MODELRUN, DAGRUN, and SCRIPTRUN

PR #436, opened 2020-07-23
Updated support email and Orobix to Tensorwork on ramp file

filipecosta90:


ramp.yml
```diff
@@ -1,6 +1,6 @@
 display_name: RedisAI
-author: Orobix and RedisLabs
-email: luca.antiga@orobix.com
+author: TensorWerk and RedisLabs
```
hhsecond, line 2:
```suggestion
author: Tensorwerk and RedisLabs
```

PR #444, opened 2020-08-13
Dependencies in ramp.yml

rafie:


ramp.yml
```diff
@@ -4,7 +4,8 @@ email: support@redislabs.com
 description: Serving tensors and executing deep learning graphs
 homepage: https://oss.redislabs.com/redisai/
 license: Redis Source Available License v1.0
-command_line_args: ""
+command_line_args: "BACKENDSPATH /var/opt/redislabs/modules/ai/{{NUMVER}}/deps"
+# command_line_args: "BACKENDSPATH /var/opt/redislabs/modules/ai/{{NUMVER}}/deps/backends"
```
filipecosta90, line 8: is this a leftover?

filipecosta90: LVGTM. only a small comment on a commented out line @rafie . Is it on purpose?

rafie: Yes, I kind of left the option of having a one-contains-all-backends .tgz (which requires tweaking of the `BACKENDSPATH`).
We can probably drop this once it hits master as a comment.

PR #446, opened 2020-08-19
Unit tests boilerplate

filipecosta90:
This PR introduces googletest as a dependency in order to properly run unit tests. It has some duplicate code that needs to be pruned prior merge.
- [x] Renaming `test` to `tests`,
- [x] Moving Python tests to `tests/flow`,
- [x] Placing unit tests under `tests/unit`.
- [x] Control flow tests via `flow/tests.sh`.
- [x] Control unit tests via `unit/tests.sh`.


test/unit/unit_tests_err.cpp
```diff
@@ -0,0 +1,73 @@
+
+#include <limits.h>
+#include "gtest/gtest.h"
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+#include "../../src/backends.h"
```
rafie, line 9: Better to have a `-I` pointing to the `../..` and then go `#include src/backends.h"` or something.

test/unit/unit_tests_err.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+
+#ifdef __cplusplus
+}
+#endif
+
+
+namespace {
+
+    void Alloc_Reset() {
```
rafie, line 23: It's ok have a 0-indent within a namespace.

test/unit/unit_tests_err.cpp
```diff
@@ -0,0 +1,73 @@
+
+#include <limits.h>
+#include "gtest/gtest.h"
+
+
+#ifdef __cplusplus
```
rafie, line 6: It's always c++, isn't it (c++ code follows)?

test/unit/unit_tests_err.cpp
filipecosta90: thank you for mentioning this Rafi. Adding the [`target_include_directories`](https://cmake.org/cmake/help/v3.12/command/target_include_directories.html#command:target_include_directories)

test/unit/unit_tests_err.cpp
filipecosta90: yes. removing it

test/unit/unit_tests_err.cpp
filipecosta90: addressed in the previous commit

rafie:
@filipecosta90 I wounder whether it will be worthwhile to separate Python tests from gtests, by:
- Renaming `test` to `tests`,
- Moving Python tests to `tests/pytest`,
- Placing `gtest` under `tests`.
We can then decide whether or not to control all tests via `tests.sh` or have a separate script for all.
I would also prefer not to use `ctest` at all if possible.

filipecosta90:
>     * `gtest`

doing so :+1:

tests/unit/tests.sh
```diff
@@ -0,0 +1,43 @@
...
+BINDIR=${BINDIR:-""}
+
+[[ ! -d $BINDIR ]] && { echo "BINDIR not provided. Aborting."; exit 1; }
+
+
+#----------------------------------------------------------------------------------------------
+
+cd $ROOT/tests/unit
+
+run_tests
```
DvirDukhan, line 43: add new line

tests/unit/unit_tests_err.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+namespace {
+
+void Alloc_Reset() {
+    RedisModule_Alloc = malloc;
+    RedisModule_Realloc = realloc;
+    RedisModule_Calloc = calloc;
+    RedisModule_Free = free;
+    RedisModule_Strdup = strdup;
+}
```
DvirDukhan, line 17: Add a new file, such as RedisGraph `rmalloc.h` that has this function in it.
Next, for each unit test create a c++ class that calls `Alloc_Reset`
```
#include "gtest.h"

#ifdef __cplusplus
extern "C" {
#endif
#include "path/to/rmalloc.h"
#ifdef __cplusplus
}
#endif

class MyTestClass:
	public ::testing::Test {
  protected:
	static void SetUpTestCase() { // Use the malloc family for allocations
		Alloc_Reset();
	}
};
```

tests/unit/unit_tests_tensor.cpp
```diff
@@ -0,0 +1,61 @@
...
+    // clear error to reuse structure
+    RAI_ClearError(err);
+
+    RAI_SetError(err,RAI_EMODELCREATE,"ERR specific error message");
+    EXPECT_STREQ(RAI_GetErrorOneLine(err),"ERR specific error message");
+    EXPECT_EQ(RAI_GetErrorCode(err),RAI_EMODELCREATE);
+
+    RAI_FreeError(err);
+}
+
```
DvirDukhan, line 60: I think you should remove this file

DvirDukhan (commented):
Look good
minor comments

tests/flow/tests_pytorch.py
```diff
@@ -253,7 +253,7 @@ def test_pytorch_modelrun_autobatch_badbatch(env):
         model_pb = f.read()
 
     ret = con.execute_command('AI.MODELSET', 'm{1}', 'TORCH', 'CPU',
-                              'BATCHSIZE', 4, 'MINBATCHSIZE', 3, 'BLOB', model_pb)
+                              'BATCHSIZE', 4, 'MINBATCHSIZE', 2, 'BLOB', model_pb)
```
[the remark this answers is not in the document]
filipecosta90: might have slipped the fingers on a merge conflict. Fixing it :)

tests/unit/unit_tests_err.cpp
filipecosta90: thank you @DvirDukhan . proceed with the changes.

tests/unit/unit_tests_tensor.cpp
filipecosta90: removed

DvirDukhan (approved):

PR #450, opened 2020-08-24
v1.0 cherry picks

filipecosta90:
Currently cherry picked, but still not ready to be merged due to failed tests (revising and fixing it):
> ba008bbc6b31c1a93272e5abf6902ffb577b0db7 Build: setuptools-related fix (#455)
> c3b3e7c93cd5391f880db4e32cff53ab633335e2 Fixed artifacts handling + added tests logs aggregation (#445)
> b1494aa Fixed flagged as "getkeys-api" during the registration ( AI.DAGRUN, AI.DAGRUN_RO, AI.MODELRUN, AI.SCRIPTRUN ) (#438)
> 86ec431 Dependencies in ramp.yml (#444)
> 573d7dd Fixed platforms build problem (#441)
> 26faf7e CircleCI: Fixed problem with GPU testing (#440)
> 3019629 Updated support email and Orobix to Tensorwork on ramp file (#436)
> 544e36e [add] Add relevant RedisAI config entries to the INFO output, so that standard monitoring systems would be able to monitor them. (#396)
> 6a14c04 Update mkdocs.yml (#432)
> f124a0e submodule pull moved up (#428)
> 66477b4 Docker6.0.5 (#427)
> 77e94a0 Added hooks/post_checkout to handle dockerhub build failures (#425)
> 9cb7791 CircleCI: fixed redis installation in coverage builds (#423)
> aa332b5 Add missing files
> cb70069 Add test for inconsistent zero batch output
> 2904fdf Update README.md
> ccf0d5c Fix run info memleaks (#412)
> 7f87f85 YOLOv3 crash with empty image hotfix
> f7fbb29 Enable chunking for large models in AOF, RDB, replication and MODELGET (#387)
> a5d1524 Avoid splitting outputs in batches when nbatches == 1 (#406)
> fcdf0d2 Llapi updates (#400)
> 6e65a42 CI: platform-build-defs -> on-master-and-version-tags (#405)
> 393604a CircleCI: increased GPU test timeout to 40m (#404)
> ad6150f Multi-platform build (#398)
> 6df468f atomic ref count (#403)
> 1a3f394 Add support for variadic arguments to SCRIPT (#395)
> 3cd5caf [WIP] Enable AI.SCRIPTRUN on AI.DAGRUN* (#383)
> a742faa [WIP] add C API reference to docs using auto tool  (#368)
> 2d2a0af Fix for python3 issues in readies (#379)
> 16729d3 Introduced readies submodule (#377)

PR #451, opened 2020-08-26
changed encver compare

DvirDukhan:


lantiga:
This enables chunking if the version is `= 100`, although chunking is being backported in the context of #450. In other words, chunking wasn't there in RedisAI 1.0. What is the rationale for this change?

DvirDukhan:
@lantiga 
`1.0.x` encver is 100
this means that backporting chunking to `1.0.2` is will cause it to encode the model in chunks but decode it in one string, which causes an invalid read of string, while the actual value itself is the number of chunks.
So perhaps a better way is to disable chunking in `1.0.x`? We can drop the backport to 1.0.x (better) or have 2 encoders (I prefer not).

lantiga:
I see. I propose we could release proper chunking together with parallel DAG in a `1.1.0` release, while leaving chunking out for `1.0.x` (including the current backport.

DvirDukhan:
Closing this PR

PR #454, opened 2020-08-31
Increase refcount for persisted tensors

lantiga:
This PR is a hotfix for a memory issue with persisted tensor on master.

The issue has been spotted during the analysis of #450.


lantiga:
@rafie FYI there's a CI issue on the gpu machine

rafie:
@lantiga pip is giving us some hard times with the 20.x versions. I'll resolve it on master.

PR #472, opened 2020-10-05
redis-cli modelget limitation

DvirDukhan:


lantiga (changes requested):
I would add a link to the issue (either ours or that is Redis) so that the implications are clearer and there is a way to track the resolution of the issue.

alonre24 (approved):

PR #489, opened 2020-11-03
Execute Redis commands in Torch Script

DvirDukhan:
This PR scope:
- [x] Add `redis` namespace resolver to torch
- [x] Add `redis.execute` custom operator to execute `RedisModule_Call` within the operator scope or from auxiliary function.
- [x] Transform `RedisModuleCallReply` to a torch representation - using `torch::IValue`
- [x] Handle redis exceptions
- [x] Memory management
- [x] Tests


src/redisai.c
```diff
@@ -578,9 +578,6 @@ int RedisAI_ScriptRun_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
         return RedisAI_ScriptRun_IsKeysPositionRequest_ReportKeys(ctx, argv, argc);
     }
 
-    if (argc < 6)
```
lantiga, line 581: We still have a minimum arity, like key and fn_name

tests/flow/test_data/redis_scripts.py
```diff
@@ -0,0 +1,66 @@
+
+def redis_string_int_to_tensor(redis_value: Any):
+    return torch.tensor(int(str(redis_value)))
```
lantiga, line 3: One thing we need to make sure is that we move the tensor to the right device. We can ask the user to do so, but it needs to be done otherwise a script running on GPU will find itself with an input on CPU and will fail.

lantiga (changes requested):
Looks good, we need to add a test in which we run SCRIPT on the GPU and the tensor built from a Redis value gets moved to the GPU.

src/redisai.c
DvirDukhan: this is checked further on the parsing function

tests/flow/test_data/redis_scripts.py
lantiga: I would just add a note here (and maybe in the docs) mentioning the device placement.
The ideal solution here would be to have a `SCRIPT_DEVICE` global to allow a user to do
```
a_tensor = torch.tensor(int(str(redis_value))).to(device=SCRIPT_DEVICE)
```
and have tensors created within the script be located on the same device as the eventual script inputs.

src/command_parser.c
```diff
@@ -183,41 +183,51 @@ static int _ScriptRunCommand_ParseArgs(RedisModuleCtx *ctx, RedisModuleString **
...
                 RAI_SetError(error, RAI_ESCRIPTRUN,
                              "ERR Already encountered a variable size list of tensors");
                 return REDISMODULE_ERR;
             }
             varidic_start_pos = ninputs;
-        } else {
+            continue;
+        }
+        // Parse argument name
+        {
```
alonre24, line 230: Why is this in a block?
And also, I think that the modified parsing logic allows to avoid INPUTS/OUTPUTS all together. For example, the following command would not raise an error (and will refer all the keys as inputs):
AI.SCRIPTRUN script_key func_name key1 key2 ...

src/command_parser.c
```diff
@@ -183,41 +183,51 @@ static int _ScriptRunCommand_ParseArgs(RedisModuleCtx *ctx, RedisModuleString **
...
+            is_output = false;
+            continue;
+        }
         if (!strcasecmp(arg_string, "OUTPUTS") && !is_output) {
             is_input = false;
             is_output = true;
-        } else if (!strcasecmp(arg_string, "$")) {
+            continue;
+        }
+        if (!strcasecmp(arg_string, "$")) {
```
alonre24, line 220: if (!strcasecmp(arg_string, "$") && is_input)
otherwise, varidic can inputs can come in the middle of OUTPUTS.

src/command_parser.c
```diff
@@ -183,41 +183,51 @@ static int _ScriptRunCommand_ParseArgs(RedisModuleCtx *ctx, RedisModuleString **
...
+        // Parse timeout arg if given and store it in timeout
+        if (!strcasecmp(arg_string, "TIMEOUT") && !timeout_set) {
+            if (_parseTimeout(argv[++argpos], error, timeout) == REDISMODULE_ERR)
+                return REDISMODULE_ERR;
+            timeout_set = true;
+            continue;
+        }
+
+        if (!strcasecmp(arg_string, "INPUTS") && !is_input) {
+            is_input = true;
```
alonre24, line 211: I think that this way, INPUTS can appear after OUTPUTS (even more than once...)
Consider add another flag like "input_done"?

alonre24 (commented):
Looks good, no comments.

alonre24 (approved):

lantiga (approved):
LGTM

PR #491, opened 2020-11-11
Libtorch num threads

hhsecond:
PR to solve #203


hhsecond:
Closing since the history is polluted. A new one is opened with the necessary changes https://github.com/RedisAI/RedisAI/pull/499

PR #499, opened 2020-11-18
Libtorch num threads

hhsecond:
PR to solve #203

### DIFF
--- a/tests/flow/tests_pytorch.py
+++ b/tests/flow/tests_pytorch.py
@@ -1,6 +1,7 @@
 import redis
 
 from includes import *
+from RLTest import Env
 
 '''
 python -m RLTest --test tests_pytorch.py --module path/to/redisai.so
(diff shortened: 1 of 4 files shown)


hhsecond:
I am not sure how to test the configuration changes, like this one. I think we should have them (for other config changes as well). Please advice!

DvirDukhan:
@hhsecond 
I believe you can achieve it by exposing 
`AI.INFO` without key parameter 
This can allow us to expose additional module-info such as module version, backend versions, etc...
@filipecosta90 did an overview of the module-info command, not a long while ago, but I don't recall what was on it, so now it is a good time to address it.

hhsecond:
@DvirDukhan If I understood you correctly, `AI.INFO` would give access to the current configurations, right?
What I meant was not just to read the configuration but to test by changing the configurations. Especially, changing the configurations such as the number of inter-op/intra-op threads

DvirDukhan:
@hhsecond
in RLTest you can set the env with `moduleargs`
```
env = Env(moduleArgs='INTER_OP 10')
```
so if you call the the `AI.INFO` during the test you will be able to assert it is in fact 3

```
env = Env(moduleArgs='INTER_OP 10')
r = env.getConnection()
res = r.executeCommand('AI.INFO')
env.assertEqual(3, res[index_of_inter_op_config])
```
Something like that

hhsecond:
@DvirDukhan amazing. I learned something new today. I'll add tests soon

lantiga (dismissed):
LGTM!

tests/flow/tests_pytorch.py
```diff
@@ -928,3 +929,26 @@ def test_pytorch_model_rdb_save_load(env):
     env.assertTrue(dtype_memory == dtype_after_rdbload)
     env.assertTrue(shape_memory == shape_after_rdbload)
     env.assertTrue(data_memory == data_after_rdbload)
+
+
+def test_parallelism():
+    env = Env(moduleArgs='INTRA_OP_PARALLELISM 1 INTER_OP_PARALLELISM 1')
```
filipecosta90, line 935: @hhsecond LGTM, but while were at it can we add @DvirDukhan recomendation on the check:
```
res = r.executeCommand('INFO MODULES')
## check for the ai section
## check for the inter and intra op values there
env.assertEqual(3, res[index_of_inter_op_config])
```

filipecosta90 (changes requested):

tests/flow/tests_pytorch.py
DvirDukhan: @hhsecond agree with @filipecosta90

hhsecond: Done

filipecosta90 (approved):

tests/flow/tests_pytorch.py
```diff
@@ -928,3 +929,36 @@ def test_pytorch_model_rdb_save_load(env):
...
+    ret = con.execute_command('AI.MODELSET', 'm{1}', 'TORCH', DEVICE, 'BLOB', model_pb)
+    ensureSlaveSynced(con, env)
+    con.execute_command('AI.MODELRUN', 'm{1}', 'INPUTS', 'a{1}', 'b{1}', 'OUTPUTS', 'c{1}')
+    ensureSlaveSynced(con, env)
+    values = con.execute_command('AI.TENSORGET', 'c{1}', 'VALUES')
+    env.assertEqual(values, [b'4', b'6', b'4', b'6'])
+    load_time_config = {k.split(":")[0]: k.split(":")[1]
+                        for k in con.execute_command("INFO MODULES").decode().split("#")[3].split()[1:]}
+    env.assertEqual(load_time_config["ai_inter_op_parallelism"], "1")
+    env.assertEqual(load_time_config["ai_intra_op_parallelism"], "1")
```
DvirDukhan, line 958: @hhsecond can you return it from `get_num_interop_threads`?
returning from the load_time_config params is no guarantee for it to be actually set

hhsecond: As discussed over slack, I'll wait for the `AI.INFO` changes you proposed - I think it is a nice addition. Merging this PR for now

PR #503, opened 2020-11-22
fix_macos_get_deps

### LINKED ISSUE #502: Installation from the source is failing in MacOS
I found a recent change of `$OS == macos` to `$OS == macosx` in the `get_deps.sh` which is causing installation failure in macos bigsur. 
Here is the platform information from `paella`
```
{'os': 'macos', 'dist': '', 'os_ver': '10.16', 'os_full_ver': '10.16', 'osnick': 'macos16', 'arch': 'x64', 'strict': False}
```

DvirDukhan:
fix #502

### DIFF
--- a/get_deps.sh
+++ b/get_deps.sh
@@ -52,7 +52,7 @@ OS=$(python3 $HERE/opt/readies/bin/platform --os)
 ARCH=$(python3 $HERE/opt/readies/bin/platform --arch)
 
 # avoid wget warnings on macOS
-[[ $OS == macosx ]] && export LC_ALL=en_US.UTF-8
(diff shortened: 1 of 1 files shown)


hhsecond:
Looks straightforward forward but I was wondering why did we change from `macos` to `macosx` in the first place. I mean, are we fixing this by breaking another version of mac?

rafie:
@hhsecond I first introduced it as `macosx`, then Apple came out with `macOS 11`, and the rest is history.

rafie (approved):

hhsecond (approved):

PR #507, opened 2020-11-23
Add valgrind to CI

DvirDukhan:


tests/flow/valgrind.sh
```diff
@@ -11,9 +11,9 @@ error() {
 # trap error ERR
 
 HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
-. $HERE/../opt/readies/shibumi/functions
```
rafie, line 14: why?

DvirDukhan: you want to start from `HERE` and not from `THERE`
https://www.youtube.com/watch?v=9m-kbBamg_U

rafie: I was wondering about why do we exclude the `shibumi/functions`, as those (for example) provide the `realpath` that's missing from macOS.
We can, of course, go `export ROOT=$(cd $HERE/../..; pwd)` and make `realpath` redundant, but the basic idea is that the shibumi thing will give us all kind of nice toys when doing bash so it's worthwhile to keep it around.

rafie (dismissed):

rafie (approved):

PR #514, opened 2020-11-26
Applied clang-format to src/* files. Failing CI if linter rules not followed

filipecosta90:
This is a maintenance PR to ensure that feature PRs are not mixed with formatting ones.
- The formatting rules are the ones present in RedisTimeSeries plus the addition of `AlignConsecutiveMacros: true` given that clang-format v10 and v11 handle that rule differently by default. 
- It enables:
  - [x] clang-format installation into system-setup.py. @rafie can you check the steps I've added per OS. 
  - [x] clang-format check into CI config.yml. To fail faster it will only start other build variants after the lint check ( in 30secs we can lint the project and fail if required vs spending 40min of compute power ) 
  - [x] clang-formatting/linting via Makefile. Added machine and docker variants, namely `format` and `format-docker`, `lint` and `lint-docker`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,6 +114,19 @@ commands:
             aws s3 cp artifacts/ s3://redismodules/$PACKAGE_NAME/ --acl public-read --recursive --exclude "*" --include "*.zip" --include "*.tgz"
 
 jobs:
+  lint:
+    docker:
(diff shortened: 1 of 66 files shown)


DvirDukhan (approved):

DvirDukhan:
@lantiga please review.

lantiga (approved):
Looks great!

Issue #518, opened 2020-11-29
Free DAG run_info based on the ref_count, instead of master field.


alonre24:
DAG run_info may have several of shallow copies that share some of this object internals. The internals should be freed in the run_info free function whenever the ref_count hits zero. The current implementation chooses to free the internals based on the master field of the copy, which is a leftover that need to be replaced.

PR #519, opened 2020-11-30
Remove master field from DAG run_info

alonre24:
#518 
Free the run_info internals if ref_count hits zero, without checking if it is the master copy or not.

### DIFF
--- a/src/run_info.c
+++ b/src/run_info.c
@@ -129,7 +129,6 @@ int RAI_InitRunInfo(RedisAI_RunInfo **result) {
     rinfo->dagDeviceOpCount = 0;
     rinfo->dagDeviceCompleteOpCount = 0;
     pthread_rwlock_init(rinfo->dagLock, NULL);
-    rinfo->master = 1;
     rinfo->timedOut = RedisModule_Calloc(1, sizeof(int));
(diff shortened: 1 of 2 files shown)


src/run_info.c
```diff
@@ -203,7 +201,7 @@ void RAI_FreeRunInfo(RedisModuleCtx *ctx, struct RedisAI_RunInfo *rinfo) {
     if (!rinfo) {
         return;
     }
-    if (rinfo->master == 0) {
+    if (*rinfo->dagRefCount == 0) {
```
lantiga, line 204: Isn't this the exact opposite?

lantiga (changes requested):

src/run_info.c
alonre24: you are right of course....
Are we testing it somewhere? This should have crash for at least two rinfo copies right? but the tests passed....

lantiga: We are testing DAGs on multiple devices, but whether they would crash if deallocated out of order it needs to be investigated. It might be that they crash once every 20 runs.

lantiga (changes requested):

lantiga (approved):

PR #520, opened 2020-11-30
Make atomic read for ref_count in DAG run_info free function.

alonre24:


lantiga (approved):
All good but there's a tab in there :-)

src/run_info.h
```diff
@@ -103,6 +103,13 @@ int RAI_InitRunInfo(RedisAI_RunInfo **result);
 
 int RAI_ShallowCopyDagRunInfo(RedisAI_RunInfo **result, RedisAI_RunInfo *src);
 
+/**
+ * Frees the shallow copy of RedisAI_RunInfo pointed by rinfo.
+ * @param rinfo copy to be freed.
+ * @retval The ref_count of the rinfo object after freeing this copy.
+ */
+long long RAI_DagRunInfoDecreaseFetch(RedisAI_RunInfo *rinfo);
```
DvirDukhan, line 111:
```suggestion
long long RAI_DagRunInfoFreeShallowCopy(RedisAI_RunInfo *rinfo);
```

Also, if it is not a big refactor, make something like
`RedisAI_RunInfo * RAI_DagRunInfoCreateShallowCopy(RedisAI_RunInfo *rinfo);`
we should avoid breaking the abstraction of the object.

src/run_info.c
```diff
@@ -197,20 +197,26 @@ void RAI_FreeDagOp(RedisModuleCtx *ctx, RAI_DagOp *dagOp) {
...
+
+void RAI_FreeRunInfo(RedisModuleCtx *ctx, struct RedisAI_RunInfo *rinfo) {
+    if (!rinfo) {
         return;
-    } else {
-        pthread_rwlock_destroy(rinfo->dagLock);
-        RedisModule_Free(rinfo->dagLock);
     }
+    long long ref_count = __atomic_load_n(rinfo->dagRefCount, __ATOMIC_RELAXED);
+    RedisModule_Assert(ref_count == 0);
```
DvirDukhan, line 217: In the current scenario, this is ok, since you assume there is one owner to the RunInfo

src/run_info.c
```diff
@@ -197,20 +197,26 @@ void RAI_FreeDagOp(RedisModuleCtx *ctx, RAI_DagOp *dagOp) {
...
+}
+
+void RAI_FreeRunInfo(RedisModuleCtx *ctx, struct RedisAI_RunInfo *rinfo) {
+    if (!rinfo) {
         return;
-    } else {
-        pthread_rwlock_destroy(rinfo->dagLock);
-        RedisModule_Free(rinfo->dagLock);
     }
+    long long ref_count = __atomic_load_n(rinfo->dagRefCount, __ATOMIC_RELAXED);
```
DvirDukhan, line 216: if the function is being called only when ref_count is 0, there is no need for atomic, since the caller probably already called free on the shallow copy and knows the ref count is 0

src/run_info.c
```diff
@@ -197,20 +197,26 @@ void RAI_FreeDagOp(RedisModuleCtx *ctx, RAI_DagOp *dagOp) {
     }
 }
 
-void RAI_FreeRunInfo(RedisModuleCtx *ctx, struct RedisAI_RunInfo *rinfo) {
-    if (!rinfo) {
-        return;
+long long RAI_DagRunInfoDecreaseFetch(RedisAI_RunInfo *rinfo) {
+    long long ref_count = __atomic_sub_fetch(rinfo->dagRefCount, 1, __ATOMIC_RELAXED);
+    RedisModule_Assert(ref_count >= 0);
```
DvirDukhan, line 202:
```suggestion
    RedisModule_Assert(ref_count >= 0 && "Tried to free the original RunInfo object");
```

src/run_info.h
alonre24: There is RAI_ShallowCopyDagRunInfo, but the ref_count increase is done outside the function (after we create n copies, we increase the ref_count by n). Should we move it to be inside the function?

lantiga: I think it's a good idea, it would have an equivalent effect

DvirDukhan (approved):

PR #523, opened 2020-12-01
Added PROFILE option to makefile. Added option to skip building unit tests

filipecosta90:
This PR eases the process of collecting profile info ( without changing CMakeLists.txt all the time ). 
`make PROFILE=1` should be enough to build in release mode, with symbols, and keeping frame pointers.  
The `-profile` suffix will be added to the build variant folder. Please notice that PROFILE option makes sense for release ( debug builds already contain this flag )


opt/Makefile
```diff
@@ -115,10 +118,17 @@ else
 USE_COVERAGE=off
 endif
 
+ifeq ($(PROFILE),1)
+PROFILE=on
+else
+PROFILE=off
+endif
+
```
rafie, line 126: Problems can arise from it. Either we use `override`, or (better) USE_PROFILE=on|off.

filipecosta90: thank you for the test fast review @rafie . addressed in the latest commit.

rafie (approved):

opt/Makefile
```diff
@@ -63,10 +65,11 @@ make docker           # build docker image
...
 fetch and build options:
-  WITH_TF=0     # SKip TensofFlow
-  WITH_TFLITE=0 # SKip TensofFlowLite
-  WITH_PT=0     # Skip PyTorch
-  WITH_ORT=0    # SKip ONNXRuntime
+  WITH_TF=0           # SKip TensofFlow
+  WITH_TFLITE=0       # SKip TensofFlowLite
+  WITH_PT=0           # Skip PyTorch
+  WITH_ORT=0          # SKip ONNXRuntime
+  WITH_UNIT_TESTS=0   # SKip ONNXRuntime
```
DvirDukhan, line 72:
```suggestion
  WITH_UNIT_TESTS=0   # SKip unit tests
```

DvirDukhan (approved):

PR #528, opened 2020-12-03
Return an error when trying to run MODELRUN/SCRIPTRUN/DAGRUN through multi or lua

alonre24:
This is a temporary solution until we support non-blocking execution of these commands, since soon redis will not support client blocking through multi/lua (it will crash).


DvirDukhan (dismissed):

lantiga (dismissed):
Looks good but I just saw there's a cluster-related failure we need to resolve (should be quick)

lantiga (approved):
LGTM

DvirDukhan (approved):

Issue #515, opened 2020-11-27
Improve RAI_TensorSet_RedisCommand memory growth/page_fault code path by reusing already allocated redis BLOB


filipecosta90:
We saw that on large tensors, for example an image with 1 x 3 x 224 x 224 ( 602 KiB ), the CPU cycles (wasted) on Allocator take a big chunk of the total redis-server main thread CPU cycles. In the example bellow we observe 42% wasted cpu cycles. 
We should investigate ways to improve it by testing reusing previously allocated command arg memory. Only possible for `BLOB` tensorsets...  
![image](https://user-images.githubusercontent.com/5832149/100437096-2e025880-3098-11eb-82f4-c4bc26400b96.png)

lantiga:
We have two options:
1. check if a key being set in TENSORSET, MODELRUN OUTPUT, SCRIPTRUN OUTPUT or DAGRUN PERSIST contains a tensor with the same buffer size (in bytes) that we can re-use (instead of deallocating / reallocating)
2. keep a memory pool for tensors where we keep tensors for a while for reuse before deallocating them

The advantage of 2) is that we could use the pool whenever we need to allocate a tensor, irrespective of whether they are associated to an actual key (e.g. TENSORSET within a DAG). The downside is that the logic for the pool needs to be thought out well, especially the logic for deallocating, to avoid having a lot of allocated memory that is not used.

PR #540, opened 2020-12-20
Reuse memory in TENSORSET

lantiga:
Addresses #515 by reusing memory allocated in `argv` in `TENSORSET`.


src/tensor.c
```diff
@@ -814,7 +868,14 @@ int RAI_parseTensorSetArgs(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
     size_t datalen;
     const char *data;
     DLDataType datatype = RAI_TensorDataTypeFromString(typestr);
-    *t = RAI_TensorCreateWithDLDataType(datatype, dims, ndims, tensorAllocMode);
+    if (datafmt == REDISAI_DATA_BLOB) {
```
DvirDukhan, line 871: since you are checking it here
the `switch` in  line 889 is redundant since it checks only a single case. I think you can move its content to the else block in line 875

DvirDukhan (commented):
looks good
small comment

src/tensor.c
lantiga: Done

filipecosta90:
I've quickly ran the benchmark that produced the best result for autobatching variation ( 60 clients, autobatching 30 ) ( command `OUTPUT_NAME_SUFIX=tensorset_PR_ MIN_CLIENTS=60 MAX_CLIENTS=60 MIN_TENSOR_BATCHSIZE=0 MAX_TENSOR_BATCHSIZE=0 MAX_BATCHSIZE=30 MIN_BATCHSIZE=30 SLEEP_BETWEEN_RUNS=0 DATABASE_HOST=10.3.0.207 DATABASE_PORT=6380 NUM_VISION_INFERENCES=25000 ./scripts/run_inference_redisai_vision.sh` ) and we see an reduction in the memory BW of around 300MB/sec. 

Notice that at ~= 400 inferences/sec we have an expected BW of 224 * 224 * 3 * 4bytes * 400 / ( 1024^3 ) = 0.23 GB/sec as show bellow by the table.


tensor size bytes | 602112
-- | --
tensorset expected BW GB/sec @400 inferences/sec | 0.2335455322

 
If we look at the numbers we notice that the memory BW reduction was of aroud 0.3 GB/sec ( matches the expected ) and the expected improvement of 5% both in the overall ops/sec and inference latency.

notes | commit | overall inferences/sec | p50 | p75 | p99 | page faults/sec | memory BW GB/sec
-- | -- | -- | -- | -- | -- | -- | --
master | e06c663c74bda3ba543b77e0f997babacd390c9c | 397.85 | 151.42 | 154.37 | 162.56 | 1400000 | 5.340576172
PR | 52c42350693e06e4f92dd28590a58b0e3781f9dd | 416.48 | 143.87 | 147.07 | 156.16 | 1313000 | 5.00869751
-- | %improvement | 4.68% | 4.99% | 4.73% | 3.94% | 6.21% | 6.21%


Detail of memory BW of e06c663c74bda3ba543b77e0f997babacd390c9c 
( the drops mean a new test iteration ( 3 iterations ) )
image

( the drops mean a new test iteration ( 3 iterations ) )
Detail of memory BW of 52c42350693e06e4f92dd28590a58b0e3781f9dd

image

----
Bottom line we see that this generally improves the inference performance ( even when it's faded by modelrun+tensorget) so I would merge this asap and work on further reducing the overall mem bandwidth by for example pushing forward the investigation of reusing tensor allocated memory on backends ( example of tensorflow with dlpack inputs )

filipecosta90 (approved):

PR #542, opened 2020-12-22
Fix modelscan crash due to tag being NULL

### LINKED ISSUE #541: RedisAI crashes on MODELSCAN & SCRIPTSCAN
The latest push to master introduce a regression that causes RedisAI to crash when doing `MODELSCAN`. Interestingly, if the `MODELSET` is tagged, it doesn't crash. I haven't looked into the core for the exact crash issue yet but made a script to reproduce if anybody else has time (run it from the root of `redisai-py` to match the relative path to the provided model)

```python
from ml2rt import load_model
from redisai import Client

con = Client()

model_path = 'test/testdata/'
ptmodel = load_model(model_path + "pt-minimal.pt")


def core_dump_when_not_tagged():
    con.modelset("pt_model", 'torch', 'cpu', ptmodel)
    con.modelscan()


def no_core_dump_when_tagged():
    con.modelset("pt_model", 'torch', 'cpu', ptmodel, tag='v1.0')
    con.modelscan()


if __name__ == "__main__":
    no_core_dump_when_tagged()
    core_dump_when_not_tagged()
```

Crash on SCRIPTSCAN is also easy to reproduce

lantiga:
Fixes #541

### DIFF
--- a/src/model.c
+++ b/src/model.c
@@ -119,7 +119,7 @@ static void *RAI_Model_RdbLoad(struct RedisModuleIO *io, int encver) {
     model->infokey =
         RAI_AddStatsEntry(stats_ctx, stats_keystr, RAI_MODEL, backend, stats_devicestr, stats_tag);
 
-    RedisModule_FreeString(stats_keystr);
+    RedisModule_FreeString(NULL, stats_keystr);
 
     return model;
 }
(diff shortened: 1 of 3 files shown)


alonre24 (approved):

lantiga:
Adding the test, don't merge yet thx

lantiga:
Done, ready to go

DvirDukhan (approved):

Issue #544, opened 2020-12-24
Unsafe DAG runInfo release upon error


alonre24:
After running a DAG op, we check if we encountered an error during the run. If an error has occurred, and DAG refCount hits zero, we call the "OnFinish" callback where we free and release the DAG runInfo. However, after the "OnFinish" is called we access the inner fields of the DAG runInfo which may lead to a crash:

            int device_complete_after_run = RedisAI_DagDeviceComplete(batch_rinfo[0]);
            int dag_complete_after_run = RedisAI_DagComplete(batch_rinfo[0]);

PR #545, opened 2020-12-24
Fix unsafe dag release upon error

alonre24:
#544
Bug fix (temporary until broader refactor).

### DIFF
--- a/src/background_workers.c
+++ b/src/background_workers.c
@@ -391,19 +391,16 @@ void *RedisAI_Run_ThreadMain(void *arg) {
 
                 // Run is over, now iterate over the run info structs in the batch
                 // and see if any error was generated
-                int dagError = 0;
+                int first_dag_error = 0;
                 for (long long i = 0; i < array_len(batch_rinfo); i++) {
                     RedisAI_RunInfo *rinfo = batch_rinfo[i];
-                    // We lock on the DAG because error could be set from
-                    // other threads operating on the same DAG (TODO: use atomic)
-                    dagError = __atomic_load_n(rinfo->dagError, __ATOMIC_RELAXED);
-
                     // We record that there was an error for later on
-                    run_error = dagError;
-
+                    run_error = __atomic_load_n(rinfo->dagError, __ATOMIC_RELAXED);
+                    if (i == 0 && run_error == 1)
+                        first_dag_error = 1;
                     // If there was an error and the reference count for the dag
                     // has gone to zero and the client is still around, we unblock
-                    if (dagError) {
+                    if (run_error) {
                         RedisAI_RunInfo *orig = rinfo->orig_copy;
                         long long dagRefCount = RAI_DagRunInfoFreeShallowCopy(rinfo);
                         if (dagRefCount == 0) {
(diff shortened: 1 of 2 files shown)


src/background_workers.c
```diff
@@ -391,19 +391,16 @@ void *RedisAI_Run_ThreadMain(void *arg) {
 
                 // Run is over, now iterate over the run info structs in the batch
                 // and see if any error was generated
-                int dagError = 0;
+                int first_dag_error = 0;
```
DvirDukhan, line 394: Is there no usage for `dagError` after this block?
also
please use `bool` or dedicated `enum` instead of `int`

src/background_workers.c
```diff
@@ -415,12 +412,17 @@ void *RedisAI_Run_ThreadMain(void *arg) {
...
+                    run_queue_len = queueLength(run_queue_info->run_queue);
+                    continue;
+                }
             }
 
             // We initialize variables where we'll store the fact hat, after the current
             // run, all ops for the device or all ops in the dag could be complete. This
             // way we can avoid placing the op back on the queue if there's nothing left
             // to do.
+            assert(run_error == 0);
```
DvirDukhan, line 425:
```suggestion
            RedisModule_Assert(run_error == 0);
```

src/background_workers.c
```diff
@@ -391,19 +391,16 @@ void *RedisAI_Run_ThreadMain(void *arg) {
...
-                    // We lock on the DAG because error could be set from
-                    // other threads operating on the same DAG (TODO: use atomic)
-                    dagError = __atomic_load_n(rinfo->dagError, __ATOMIC_RELAXED);
-
                     // We record that there was an error for later on
-                    run_error = dagError;
-
+                    run_error = __atomic_load_n(rinfo->dagError, __ATOMIC_RELAXED);
+                    if (i == 0 && run_error == 1)
+                        first_dag_error = 1;
```
DvirDukhan, line 400: one liner or open a block

tests/flow/tests_onnx.py
```diff
@@ -137,6 +137,14 @@ def test_onnx_modelrun_mnist(env):
         env.assertEqual(type(exception), redis.exceptions.ResponseError)
         env.assertEqual("model key is empty", exception.__str__())
 
+    # This error is caught after the model is sent to the backend, not in parsing like before.
+    try:
+        con.execute_command('AI.MODELRUN', 'm{1}', 'INPUTS', 'a{1}', 'a{1}', 'OUTPUTS', 'b{1}')
+    except Exception as e:
+        exception = e
+        env.assertEqual(type(exception), redis.exceptions.ResponseError)
+        env.assertEqual('Expected 1 inputs but got 2', exception.__str__())
```
DvirDukhan, line 146: `str(exception)` will not work here?

src/background_workers.c
alonre24: Our CLang formatter changes the one liners to be like this.

src/background_workers.c
alonre24: no usage... 
Changed to bool

tests/flow/tests_onnx.py
alonre24: It will, but is it matter?
we use exception.__str__() many time in this test... replace all?

DvirDukhan (approved):

PR #549, opened 2020-12-29
Fix valgrind leaks (with minor refactor)

alonre24:
- In DAG_OP init, we allocate inkeys and outkeys arrays. When we parse model run command, we allocate inkeys and outkeys arrays as well (what causes the leaks). This PR contains a temporary fix, this will be refactor soon.
- In DAG_commandParser, the following line causes a leak:
RedisModuleString *mangled_key = RedisModule_CreateStringFromString(NULL, key); 
A call for RedisModule_FreeString was added.
- The allocation of DAG_device_ops array in RunInfo_Init is redundant (it is initialized in every shallow copy). 
- The function Tensor_DataTypeStr allocates memory on the heap to store the string that represents the tensor type, and it wasn't freed in one place. This allocation is replaced with stack allocation for char array - thus we save allocations and get rid of the leaks .

### DIFF
--- a/src/DAG/dag.c
+++ b/src/DAG/dag.c
@@ -884,24 +884,3 @@ void DAG_ReplyAndUnblock(RedisAI_OnFinishCtx *ctx, void *private_data) {
     if (rinfo->client)
         RedisModule_UnblockClient(rinfo->client, rinfo);
 }
-
-void Dag_PopulateOp(RAI_DagOp *currentOp, void *rctx, RedisModuleString **inkeys,
-                    RedisModuleString **outkeys, RedisModuleString *runkey) {
-
-    if (currentOp->commandType == REDISAI_DAG_CMD_MODELRUN) {
-        currentOp->mctx = (RAI_ModelRunCtx *)rctx;
-        currentOp->devicestr = currentOp->mctx->model->devicestr;
-    } else {
-        assert(currentOp->commandType == REDISAI_DAG_CMD_SCRIPTRUN);
-        currentOp->sctx = (RAI_ScriptRunCtx *)rctx;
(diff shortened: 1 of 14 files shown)


lantiga:
Hi @alonre24 , can you provide a small description of the changes and what the leak was due to?

src/DAG/dag.c
```diff
@@ -897,6 +897,10 @@ void Dag_PopulateOp(RAI_DagOp *currentOp, void *rctx, RedisModuleString **inkeys
         currentOp->devicestr = currentOp->sctx->script->devicestr;
     }
 
+    // todo: temporary patch to fix leak, need refactor
+    array_free(currentOp->inkeys);
+    array_free(currentOp->outkeys);
```
DvirDukhan, line 902: send `currentOp` inkeys and outkeys to the parsing instead of freeing them and reset the values here. do for both script and model

src/command_parser.c
```diff
@@ -309,17 +313,18 @@ int ParseScriptRunCommand(RedisAI_RunInfo *rinfo, RedisModuleCtx *ctx, RedisModu
...
     currentOp->commandType = REDISAI_DAG_CMD_SCRIPTRUN;
     Dag_PopulateOp(currentOp, sctx, inkeys, outkeys, runkey);
     rinfo->dagOps = array_append(rinfo->dagOps, currentOp);
     return REDISMODULE_OK;
 
 cleanup:
+    if (error.detail) {
+        RedisModule_Free(error.detail);
+        RedisModule_Free(error.detail_oneline);
+    }
```
DvirDukhan, line 327: clear error

lantiga (approved):
LGTM, good catches

DvirDukhan (approved):

PR #552, opened 2021-01-03
expose model inputs and outputs with respect to model definition

### LINKED ISSUE #529: Expose model inputs and outputs number in `AI.MODELGET`
As `INPUTS` and `OUTPUTS` are required only for TF models in the `AI.MODELSET` command, we want to allow investigate model's input and outputs:

- [ ] ONNX:
```
 size_t n_input_nodes;
 ort->SessionGetInputCount(session, &n_input_nodes);
 size_t n_output_nodes;
 ort->SessionGetOutputCount(session, &n_output_nodes);
```

- [ ] Torch - User defined

DvirDukhan:
closes #529


src/command_parser.c
```diff
@@ -71,15 +71,13 @@ static int _ModelRunCommand_ParseArgs(RedisModuleCtx *ctx, RedisModuleString **a
     }
     if ((*model)->inputs && (*model)->ninputs != ninputs) {
```
alonre24, line 72: Is the condition (*model)->inputs is still needed? The inputs and outputs arrays should be allocated in all backends now (not only in TF) right?

lantiga:
Looks great, the failing tests should be just a matter of updating the tests themselves.
I would remove the condition as @alonre24 pointed out.

lantiga (approved):
LGTM

alonre24 (approved):

PR #561, opened 2021-01-14
Let TENSORGET command return BLOB+META by default

alonre24:


tests/flow/tests_common.py
```diff
@@ -205,6 +205,13 @@ def test_common_tensorget(env):
         env.assertEqual(datatype.encode('utf-8'), tensor_dtype)
         env.assertEqual([2], tensor_dim)
 
+        # Confirm that default reply format is META BLOB
+        _, tensor_dtype_default, _, tensor_dim_default, _, tensor_blob_default = con.execute_command('AI.TENSORGET',
+                                                                                                     'tensor_{0}'.format(datatype))
+        env.assertEqual(tensor_dtype, tensor_dtype_default)
+        env.assertEqual(tensor_dim, tensor_dim_default)
+        env.assertEqual(tensor_blob, tensor_blob_default)
+
```
DvirDukhan, line 214: move this to a different `for` loop or different test function

src/tensor.c
```diff
@@ -851,6 +851,8 @@ uint ParseTensorGetArgs(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
         RedisModule_WrongArity(ctx);
         return fmt;
     }
+    if (argc == 2)
+        fmt = TENSOR_BLOB | TENSOR_META;
```
DvirDukhan, line 855:
```suggestion
    if (argc == 2) return TENSOR_BLOB | TENSOR_META;
```
also, one-liner or open scope after line drop

docs/commands.md
```diff
@@ -79,7 +79,7 @@ Depending on the specified reply format:
     1. The tensor's shape as an Array consisting of an item per dimension
  * **BLOB**: the tensor's binary data as a String. If used together with the **META** option, the binary data string will put after the metadata in the array reply.
  * **VALUES**: Array containing the numerical representation of the tensor's data. If used together with the **META** option, the binary data string will put after the metadata in the array reply.
-
+ * Default: META+BLOB are returned by default, in case that non of the arguments above is specified.
```
DvirDukhan, line 82:
```suggestion
 * Default: **META** and **BLOB** are returned by default, in case that non of the arguments above is specified.  
```

DvirDukhan (approved):

PR #567, opened 2021-01-19
CircleCI: support for intergration branches

rafie:


.circleci/config.yml
```diff
@@ -345,17 +352,26 @@ on-version-tags: &on-version-tags
       only: /^v[0-9].*/
 
 on-master-and-version-tags: &on-master-and-version-tags
+  filters:
+    branches:
+      only: master
+    tags:
+      only: /^v[0-9].*/
```
DvirDukhan, line 359: remove `on-master-and-version-tags` as it is no longer in use, or combine both definitions

.circleci/config.yml
```diff
@@ -405,7 +421,7 @@ workflows:
           # <<: *on-version-tags
       - deploy-snapshot:
           <<: *after-platform-builds
-          <<: *on-master
+          <<: *on-integ-branch
```
DvirDukhan, line 424: why do we want `integ` branch snapshots? will it overwrite existing versions?

rafie: like we get a master snapshot, we may want integration branch snapshots, with corresponding names, like `redisai/snapshots/redisai-cpu-onnxruntime.linux-bionic-x64.1.2.tgz`.
we can limit ourselves to `on-master` at the moment.

.circleci/config.yml
rafie: yes.

DvirDukhan (approved):

PR #581, opened 2021-01-28
tensor len member

### LINKED ISSUE #495: Improve RAI_TensorLength performance
We saw an substantial amount of cpu cycles being spent on this method, that is called for multiple reasons.
we're accessing the dl_tensor structure, and it's shape array and calling the for loop.... just to do the simple computation of multiplying the dimension sizes. 
we can have the pre-computed length without always calculating it... 
Changes will be validated with a proper microbenchmark... 
https://github.com/RedisAI/RedisAI/blob/master/src/tensor.c#L460

DvirDukhan:
this PR adds a new `len` member to `RAI_Tensor` which is lazy evaluated on `RAI_TensorLength` 
This should reduce the amount of calls to `RAI_TensorLength`

closes #495

### DIFF
--- a/src/backends/onnxruntime.c
+++ b/src/backends/onnxruntime.c
@@ -174,7 +174,7 @@ RAI_Tensor *RAI_TensorCreateFromOrtValue(OrtValue *v, size_t batch_offset, long 
         return NULL;
     }
 
-    ret = RedisModule_Calloc(1, sizeof(*ret));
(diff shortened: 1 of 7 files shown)


filipecosta90 (dismissed):

DvirDukhan:
@filipecosta90 Can we benchmark before merge?

lantiga (approved):
LGTM

PR #583, opened 2021-01-30
CircleCI: dockerhub builds

rafie:


.circleci/config.yml
```diff
@@ -422,31 +403,40 @@ on-integ-and-version-tags: &on-integ-and-version-tags
...
           <<: *on-any-branch
-      - build-debian:
+      - build-and-test:
           <<: *on-any-branch
           <<: *after-linter
+      - dockerhub-build:
+          name: dockerhub-edge
+          build_args: EDGE=1
+          <<: *after-build-and-test
+          <<: *on-any-branch
```
DvirDukhan, line 439:
```suggestion
          <<: *on-master
```

.circleci/config.yml
```diff
@@ -106,6 +106,29 @@ commands:
...
+      - checkout
+      - run:
+          name: Setup
+          command: |
+            git submodule update --init --recursive
+            ./opt/readies/bin/getpy3
+      - run:
+          name: Build for platform
+          command: |
+            #@@ docker login -u redisai -p $DOCKER_REDISAI_PWD
```
DvirDukhan, line 124: comment?

.circleci/config.yml
```diff
@@ -106,6 +106,29 @@ commands:
...
+            cd opt/build/docker
+            #@@ make build publish $(./version-params) CPU=1 OSNICK=bionic X64=1 REPO=redisai VERBOSE=1 <<parameters.build_args>>
+            #@@ make build publish $(./version-params) GPU=1 OSNICK=bionic X64=1 REPO=redisai VERBOSE=1 <<parameters.build_args>>
```
DvirDukhan, line 127: comment?

DvirDukhan (commented):
thanks @rafie !
1. edge is only on master
2. where is the logic between tagging a `1.0.x` to `1.2.x` with respect to the `latest` tag?

.circleci/config.yml
rafie: (1) this is for testing the PR. will change to "on-master" once it is working.
(2) true. will add support for `LATEST_BRANCH=1.2`

.circleci/config.yml
rafie: Don't yet have password for `dockerhub/redisai` so publish will fail.

DvirDukhan (approved):

PR #584, opened 2021-02-02
Install redis-gears from S3 as part of module deps, and test AI-Gears LLAPI.

alonre24:
Download RedisGears from S3 as part of flow tests dependencies, and add tests that uses gears through the LLAPI to test suit.


tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,31 @@
+if [ -z ${OS+x} ]
+then
+      echo "os is not set"
+      exit 1
+fi
```
rafie, line 5: Please replace with:
```
#!/bin/bash

HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
. $HERE/readies/shibumi/functions
ROOT=$(realpath $HERE/../..)
READIES=$ROOT/opt/readies/bin

OS=$($READIES/bin/platform --osnick)
```

Also need to verify file permissions.

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo "installing redisgears for :" $OS
```
rafie, line 7: `echo "Installing redisgears for: $OS"`

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,31 @@
...
+
+WORK_DIR=./bin/RedisGears/
```
rafie, line 9:
```
cd $ROOT
WORK_DIR=bin/$($READIES/bin/platform -t)/RedisGears
```

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,31 @@
...
+REDISGEARS_ZIP=redisgears.linux-$OS-x64.master.zip
+REDISGEARS_DEPS=redisgears-python.linux-$OS-x64.master.tgz
+REDISGEARS_S3_PATH=http://redismodules.s3.amazonaws.com/redisgears/snapshots/$REDISGEARS_ZIP
+REDISGEARS_DEPS_S3_PATH=http://redismodules.s3.amazonaws.com/redisgears/snapshots/$REDISGEARS_DEPS
+
+mkdir -p $WORK_DIR
+
+if [ -f "$WORK_DIR$REDISGEARS_ZIP" ]; then
```
rafie, line 17:
```
if [[ -f $WORK_DIR/$REDISGEARS_ZIP ]]; then
```

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,31 @@
...
+    echo "Skiping RedisGears download"
+else
+    echo "Download RedisGears"
+    wget -P $WORK_DIR $REDISGEARS_S3_PATH
+    unzip $WORK_DIR$REDISGEARS_ZIP -d $WORK_DIR
```
rafie, line 22:
```
wget -q -P $WORK_DIR $REDISGEARS_S3_PATH
unzip -q -d $WORK_DIR $WORK_DIR/$REDISGEARS_ZIP
```

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+if [ -f "$WORK_DIR$REDISGEARS_DEPS" ]; then
+    echo "Skiping RedisGears download"
+else
+    echo "Download RedisGears deps"
+    wget -P $WORK_DIR $REDISGEARS_DEPS_S3_PATH
+    tar -C $WORK_DIR -xvf $WORK_DIR$REDISGEARS_DEPS
+fi
```
rafie, line 31:
```
if [[ -f $WORK_DIR/$REDISGEARS_DEPS ]]; then
    echo "Skiping RedisGears download"
else
    echo "Download RedisGears deps"
    wget -q -P $WORK_DIR $REDISGEARS_DEPS_S3_PATH
    tar -C $WORK_DIR -xf $WORK_DIR/$REDISGEARS_DEPS
fi
```

tests/flow/tests.sh
```diff
@@ -52,6 +52,12 @@ install_git_lfs() {
 
 #----------------------------------------------------------------------------------------------
 
+install_redis_gears() {
+
+  OS=bionic /bin/bash $HERE/Install_RedisGears.sh
```
rafie, line 57:
```
$HERE/Install_RedisGears.sh
```

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,31 @@
...
+then
+      echo "os is not set"
+      exit 1
+fi
+
+echo "installing redisgears for :" $OS
+
+WORK_DIR=./bin/RedisGears/
+REDISGEARS_ZIP=redisgears.linux-$OS-x64.master.zip
+REDISGEARS_DEPS=redisgears-python.linux-$OS-x64.master.tgz
```
rafie, line 11:
```
REDISGEARS_ZIP=redisgears.$PLAT.master.zip
REDISGEARS_DEPS=redisgears-python.$PLAT.master.tgz
```

tests/flow/tests_withGears.py
```diff
@@ -0,0 +1,326 @@
...
+
+    con.execute_command('AI.TENSORSET', 'b{1}', 'FLOAT',
+                        2, 2, 'VALUES', 2, 3, 2, 3)
+    test_data_path = os.path.join(os.path.dirname(__file__), 'test_data')
+    model_filename = os.path.join(test_data_path, 'graph.pb')
+
+    with open(model_filename, 'rb') as f:
+        model_pb = f.read()
+    ret = con.execute_command('AI.MODELSET', 'm{1}', 'TF', DEVICE,
+                              'INPUTS', 'a', 'b', 'OUTPUTS', 'mul', 'BLOB', model_pb)
```
DvirDukhan, line 295:
```suggestion
                              'INPUTS', 'a', 'b', 'OUTPUTS', 'c', 'BLOB', model_pb)
```

tests/flow/tests_withGears.py
```diff
@@ -0,0 +1,326 @@
...
+    env.assertEqual(ret[0], b'test3_OK')
+
+    values = con.execute_command('AI.TENSORGET', 'test3_res{1}', 'VALUES')
+    env.assertEqual(values, [b'4', b'6', b'4', b'6'])
+
+    ret = con.execute_command('rg.trigger', 'DAGRun_test4')
+    # This should raise an exception
+
+    error_string = b'attempted to get undefined function no_func'
+    env.assertEqual(str(ret[0])[:len(error_string)+2]+"'", "{}".format(error_string))
```
DvirDukhan, line 319: what is this?

tata406 (approved):

tata406 (commented):
Su

tata406 (approved):

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+ROOT=$(cd $HERE/../..; pwd)
```
chayim, line 4: ROOT=$(cd $HERE/../../ && pwd)

By combining them with a semi-colon (;) instead of double-ampersand (&&) your command will succeed, even if the directory fails to exist. This could have unintended consequences in many cases.

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,95 @@
...
+	else
+		eprint "Cannot match binary artifacts - build RedisGears and set GEARS_PATH"
+		exit 1
+	fi
+	platform="${os}-${nick}-${arch}"
+fi
+
+GEARS_S3_URL=http://redismodules.s3.amazonaws.com/redisgears/snapshots
+GEARS_MOD=redisgears.$platform.master.zip
+GEARS_DEPS=redisgears-python.$platform.master.tgz
```
chayim, line 60: GEARS_DEPS=redisgears-python.${platform}.master.zip
GEARS_MOD=redisgears.${platform}.master.zip

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,95 @@
...
+
+FINAL_WORK_DIR=$ROOT/bin/$($READIES/bin/platform -t)/RedisGears
```
chayim, line 62: Since line 31 has a call to platform -t, what about putting all of those together in one place, then you can refer to the variable here.

tests/flow/includes.py
```diff
@@ -10,12 +10,11 @@
...
-try:
-    sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../deps/readies"))
-    import paella
-except:
-    pass
 
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../opt/readies"))
+import paella
+
+ROOT = os.environ.get("ROOT")
```
chayim, line 17: ROOT = os.environ.get("ROOT", None)
if ROOT is None:
    sys.stderr.write("ROOT was not defined in the environment.\n")
    sys.exit(1)

tests/flow/tests_withGears.py
```diff
@@ -0,0 +1,311 @@
...
+        con = env.getConnection()
+        modules = con.execute_command("MODULE", "LIST")
+        if b'rg' in [module[1] for module in modules]:
+            return f(env, *args, **kwargs)
+        try:
+            platform = paella.Platform()
+            redisgears_dir = "{ROOT}/bin/{PLATFORM}/RedisGears".format(ROOT=ROOT, PLATFORM=platform.triplet())
+            redisgears_path = os.path.join(redisgears_dir, 'redisgears.so')
+            python_plugin_path = os.path.join(redisgears_dir, 'plugin/gears_python.so')
+            ret = con.execute_command('MODULE', 'LOAD', redisgears_path, 'Plugin', python_plugin_path, 'CreateVenv',
```
chayim, line 20: It would be better to wrap the _try_ around the execute_command line, since that's what can fail. 

Defensively, I might check for the redisgears_path and python_plugin_path (if os.path.isdir(<thing>) prior to executing the command. Then we can exit with meaningful text about missing directories.

chayim (changes requested):

tests/flow/tests_withGears.py
alonre24: An assertion that the error message is the one we expect to. Actually the error message is much longer and detailed, so we only verify that it starts with error_string. I changed line 319 so that the code will be more understandable (but it does the same)

tests/flow/Install_RedisGears.sh
```diff
@@ -0,0 +1,95 @@
...
+
+os=$($READIES/bin/platform --os)
+arch=$($READIES/bin/platform --arch)
+
+if [[ ! -z $GEARS_PATH ]]; then
+	platform=$($READIES/bin/platform -t)
+else
+	if [[ $os != linux || $arch != x64 ]]; then
+		eprint "Cannot match binary artifacts - build RedisGears and set GEARS_PATH"
+		exit 1
```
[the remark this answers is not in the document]
rafie: @DvirDukhan We currently only have prebuild Gears binaries for linux-x64. On other platforms, one needs to build Gears and point to the repo. Practically, it is easier just to skip the test altogether.

DvirDukhan (approved):

chayim (changes requested):
Think about quoting all shall variables in Install_RedisGears.sh.

This StackOverflow has a good discussion of the security implications as to why.

https://unix.stackexchange.com/questions/171346/security-implications-of-forgetting-to-quote-a-variable-in-bash-posix-shells

chayim (approved):

PR #592, opened 2021-02-03
join threads on tests

DvirDukhan:


tests/flow/tests_dag.py
```diff
@@ -771,6 +771,7 @@ def run():
             # assert that result tensor exists
             ret = con.execute_command("EXISTS {}".format(result_tensor_keyname))
             env.assertEqual(ret, 1)
+            t.join()
```
alonre24, line 774: Don't you want to join even before(line 763)? Is there a way to verify that batching actually occurred?

tests/flow/tests_onnx.py
```diff
@@ -229,6 +229,7 @@ def run():
     argmax = max(range(len(values)), key=lambda i: values[i])
 
     env.assertEqual(argmax, 1)
+    t.join()
```
alonre24, line 232: Same question here

tests/flow/tests_pytorch.py
```diff
@@ -284,6 +284,7 @@ def run():
         exception = e
         env.assertEqual(type(exception), redis.exceptions.ResponseError)
         env.assertEqual("Model did not generate the expected batch size", exception.__str__())
+    t.join()
```
alonre24, line 287: I would add "env.assertTrue(False)" after lines 274 and 282, otherwise the test doesn't guarantee that those exceptions were raised...

lantiga (dismissed):

tests/flow/tests_pytorch.py
DvirDukhan:
```python
ret = con.execute_command('AI.MODELRUN', 'm{1}', 'INPUTS', 'd{1}', 'e{1}', 'OUTPUTS', 'f1{1}', 'f2{1}')
env.assertEqual(ret, b'OK')
```
as a single call might succeed, depends on the threads timing

alonre24 (approved):

PR #619, opened 2021-02-25
Onnx runtime set allocator

alonre24:
Introduce an API that allows setting a custom allocator for onnxruntime backend. We register RedisModule Alloc and Free functions as onnx allocator, so that onnx is now using them in model create, delete and run.
The memory usage of onnx (via redis allocator) and the number of accesses to the allocator is saved as part of the module info (defined in redisai.c), under a new section named "ai_memory_info". These are exported to RedisAI as new fields in the RAI_LoadedBackend struct.


chayim (approved):

opt/build/onnxruntime/Dockerfile.arm
```diff
@@ -48,6 +48,6 @@ RUN set -e ;\
 	./build.sh ${BUILDARGS} --build_shared_lib
 # RUN ./build.sh ${BUILDARGS} --enable_pybind --build_wheel
 
-RUN ./pack.sh ${ONNXRUNTIME_VER}
+RUN ./pack.sh ${ONNXRUNTIME_VER} ${ARCH}
```
rafie, line 51: `pack.sh` can detect ${ARCH} by itself, no need to pass arg.

opt/build/onnxruntime/Dockerfile.arm7
```diff
@@ -41,6 +42,6 @@ RUN set -e ;\
 	./build.sh ${BUILDARGS} --build_shared_lib
 # RUN ./build.sh ${BUILDARGS} --enable_pybind --build_wheel
 
-RUN ./pack.sh ${ONNXRUNTIME_VER}
+RUN ./pack.sh ${ONNXRUNTIME_VER} ${ARCH}
```
rafie, line 45: `pack.sh` can detect ${ARCH} by itself, no need to pass arg.

opt/build/onnxruntime/pack.sh
```diff
@@ -2,6 +2,7 @@
 
 set -e
 VER="$1"
+PLATFORM="$2"
```
rafie, line 5: `PLATFORM=$($READIES/bin/platform -t)` should be used here.
And `linux-${PLATFORM}` should be replaced with `${PLATFORM}`.

rafie (changes requested):
`tests/module/Makefile` should be replaced by a corresponding `CMakeLists.txt`.

src/backends/onnxruntime.c
```diff
@@ -4,19 +4,99 @@
 #include "backends/util.h"
 #include "tensor.h"
 #include "util/arr_rm_alloc.h"
+#include <stdatomic.h>
 
 #include "onnxruntime_c_api.h"
 
+#define ONNX_API(x)                                                                                \
```
DvirDukhan, line 11:
```suggestion
#define ONNX_VALIDATE_STATUS(x)                                                                                \
```

DvirDukhan, line 11: add comments

src/backends/onnxruntime.c
```diff
@@ -4,19 +4,99 @@
...
+    if ((status = (x)) != NULL)                                                                    \
+        goto error;
+
+#ifdef RAI_ONNXRUNTIME_USE_CUDA
+#define GET_GLOBAL_ALLOCATOR ONNX_API(ort->GetAllocatorWithDefaultOptions(&global_allocator))
+#else
+#define GET_GLOBAL_ALLOCATOR
+#endif
```
DvirDukhan, line 19: explain this important macro

src/backends/onnxruntime.c
```diff
@@ -4,19 +4,99 @@
...
+    size_t allocated_size = RedisModule_MallocSize(p);
+    atomic_fetch_add(&OnnxMemory, allocated_size);
+    atomic_fetch_add(&OnnxMemoryAccessCounter, 1);
+    return p;
+}
+
+void AllocatorFree(OrtAllocator *ptr, void *p) {
+    (void)ptr;
+    size_t allocated_size = RedisModule_MallocSize(p);
+    atomic_fetch_add(&OnnxMemory, -1 * allocated_size);
```
DvirDukhan, line 48:
```suggestion
    atomic_fetch_sub (&OnnxMemory, -1 * allocated_size);
```

src/backends/onnxruntime.c
```diff
@@ -4,19 +4,99 @@
...
+    atomic_fetch_add(&OnnxMemoryAccessCounter, 1);
```
DvirDukhan, line 49:
```suggestion
    atomic_fetch_sub (&OnnxMemoryAccessCounter, 1);
```

rafie:
Instead of `test/module/Makefile`, better to use a `CMakeLists.txt`.
Take a look at this one, for example: https://github.com/RediSearch/RediSearch/blob/tests-with-.so/tests/so_tests/CMakeLists.txt
Should be very similar to that.

DvirDukhan (approved):

rafie (approved):

PR #638, opened 2021-03-14
changed minbatch logic

DvirDukhan:
this PR changes this `minbatch` logic check for not quitting the batching loop when `minbatchsize` is 0.
This will mimic the 1.0 behavior, as `minbatchsize` is checked outside the batching loop.

### DIFF
--- a/src/background_workers.c
+++ b/src/background_workers.c
@@ -241,7 +241,7 @@ static RedisAI_RunInfo **_BGThread_BatchOperations(RunQueueInfo *run_queue_info,
         // If batch size exceeds the minimum batch size already, then we're done.
         // Otherwise, if minbatchsize was set and the size wasn't reached,
         // loop until there's something new on the queue
-        if (minbatchsize !=0 && current_batchsize >= minbatchsize) {
+        if (minbatchsize != 0 && current_batchsize >= minbatchsize) {
             break;
         }
 
--- a/src/background_workers.c
+++ b/src/background_workers.c
@@ -238,13 +238,6 @@ static RedisAI_RunInfo **_BGThread_BatchOperations(RunQueueInfo *run_queue_info,
             continue;
         }
 
-        // If the new batch size would exceed the prescribed batch
-        // size, then quit searching.
-        // Here we could consider searching further down the queue.
-        if (current_batchsize + next_batchsize > batchsize) {
-            break;
-        }
-
         // If all previous checks pass, then keep track of the item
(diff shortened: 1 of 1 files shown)


src/background_workers.c
```diff
@@ -238,11 +257,10 @@ static RedisAI_RunInfo **_BGThread_BatchOperations(RunQueueInfo *run_queue_info,
         // there's anything else to batch
         current_batchsize += next_batchsize;
 
-        // If minbatchsize hasn't been set, or if the current batch
-        // size exceeds the minimum batch size already, then we're done.
+        // If batch size exceeds the minimum batch size already, then we're done.
         // Otherwise, if minbatchsize was set and the size wasn't reached,
         // loop until there's something new on the queue
-        if (minbatchsize == 0 || current_batchsize >= minbatchsize) {
+        if (minbatchsize != 0 && current_batchsize >= minbatchsize) {
```
alonre24, line 263: Do we really need to check it? why stopping at minbatchsize if we can extend the batch even further (untill it reaches batchsize)?

src/background_workers.c
```diff
@@ -188,11 +189,29 @@ static RedisAI_RunInfo **_BGThread_BatchOperations(RunQueueInfo *run_queue_info,
         return batch_rinfo;
     }
 
+    // Set the batch to be ready by default (optimistic), change it during run.
+    *batchReady = true;
+    bool timeout = false;
+    // If minbatchsize has been set and we are not past it, we check
+    // if the timeout for min batch has expired, in which case we proceed
+    // anyway
+    if (minbatchsize > 0 && minbatchtimeout > 0) {
```
alonre24, line 198: Maybe we can move the timeout checking to the beginning of the loop (line 214) and avoid the duplication in lines 290-278

alonre24 (commented):
Nice! Added 2 small comments for your review

src/background_workers.c
DvirDukhan: if the queue is empty, you need to see that the pending task is not timedout

src/background_workers.c
DvirDukhan: good point

alonre24 (approved):

AvitalFineRedis:
After the changes in the current PR, we saw that the previous regression (21%) was gone:
image
image

* as we focused on CPU/auto-batching, we should discuss if further tests should be made on GPU and tensor-batching.

PR #688, opened 2021-04-18
tflite build changes

### LINKED ISSUE #631: update tflite

### LINKED ISSUE #686: add to docs - build backends from source on different platforms

chayim:
Updating RedisAI to the latest version of tflite. tflite is built on the Jetson devices, as well as x86, as part of this series of changes. For now, as agreed we've removed OSX support - it will be added back in another card.


get_deps.sh
```diff
@@ -156,26 +156,24 @@ if [[ $WITH_TFLITE != 0 ]]; then
 		LIBTF_URL_BASE=https://s3.amazonaws.com/redismodules/tensorflow
 		if [[ $OS == linux ]]; then
 			TFLITE_OS="linux"
-			# if [[ $GPU != 1 ]]; then
-			# 	TFLITE_BUILD="cpu"
-			# else
-			# 	TFLITE_BUILD="gpu"
-			# fi
+			 if [[ $GPU != 1 ]]; then
+			 	TFLITE_PLATFORM="cpu"
```
rafie, line 160: Wouldn't `TFLITE_DEVICE` be more appropriate here?

opt/build/tflite/Dockerfile.x64
```diff
@@ -1,29 +1,34 @@
+ARG BAZEL_VERSION=3.1.0
+ARG TFLITE_ARCH=x86_64
 
-ARG OS=debian:buster
+ARG OS=redisfab/ubuntu1804-${TFLITE_ARCH}-bazel${BAZEL_VERSION}
```
rafie, line 4: Which container is that? ubuntu:bionic with bazel of a particular version for the given architecture? Notice that Bazel versions vary frequently, and we might not match a requirement for a specific one.

chayim: This is a container built, using the automata repo and pushed to redisfab. It specifically has the bazel version required for tflite as there's a hard requirement in the build (i.e 3.1.0).

get_deps.sh
chayim: I'd argue it's a nit. In RedisAI we use PLATFORM to mean the backend - so this is analogous. However, I don't have a strong feeling about it. @DvirDukhan do you care?

DvirDukhan:
@chayim 
which cuda version is used to build the TFLITE
I'm running my env with cuda 11.0, and I can't load the delegate

chayim:
> @chayim
> which cuda version is used to build the TFLITE
> I'm running my env with cuda 11.0, and I can't load the delegate

On Jetson it was the local 10.2. On x86-64 I _think_ 10.2 as well.

PR #707, opened 2021-04-27
docker build - change buster to bionic

alonre24:
- need to see if there are more places that need to be changed


chayim (changes requested):
Also edit opt/build/docker/Makefile (see line 7).  You're going to also want to edit Dockerfile.gpu at the root of the repo.

It's best to make these consistent - even if there's variable use.

alonre24:
Rafi updates the redisfab image instead - closed.

PR #708, opened 2021-04-28
machine change from gpu with comment as to why

chayim:


.circleci/config.yml
```diff
@@ -198,11 +198,13 @@ jobs:
...
 
+  # this build runs on a fixed machine, due to a storage need
+  # nothing about it necessitates the machine itself, other than the need for more disk.
   platforms-build:
     machine:
       enabled: true
       docker_layer_caching: true
-      resource_class: gpu.nvidia.small
+      resource_class: small
       image: ubuntu-1604-cuda-11.1:202012-01
```
rafie, line 208: We need a non-cuda image, like `ubuntu-2004:202010-01`.

chayim: I was hoping for an 1804 machine since we've tested outside of that. But.. updating.For the curious, the machine images are defined [here](https://circleci.com/docs/2.0/configuration-reference/#available-machine-images)

DvirDukhan (approved):

rafie (approved):

PR #718, opened 2021-05-03
Onnx add disable external initializers config for rce

alonre24:
Add a load time config that indicates if RedisAI runs in RCE. 
Add a boolean field to model_struct opts, disable_external_data, and set it to true if we run on RCE.
When creating a new onnx model, set the matching session to disable external initializers, and fail if the model has external initializers.
Use the new forked version of ONNX (it is 1.7.2 with the custom allocator and the option to disable extrnal initializers) as backend.


alonre24:
Closing for now, will wait for onnx new release and then compile ONNX with the DISABLE_EXTERNAL_INITIALIZERS flag. After we can use this test

alonre24:
Closing for now, will wait for onnx new release and then compile ONNX with the DISABLE_EXTERNAL_INITIALIZERS flag. After we can use this test

PR #720, opened 2021-05-04
ci fix test - removing remote docker

chayim:


.circleci/config.yml
```diff
@@ -40,8 +40,8 @@ commands:
             sudo apt-get -qq update
             sudo apt-get -q install -y git openssh-client curl ca-certificates make tar gzip
             bash <(curl -fsSL https://get.docker.com)
-      - setup_remote_docker:
-          docker_layer_caching: true
+#      - setup_remote_docker:
+#          docker_layer_caching: true
```
DvirDukhan, line 44: Are you sure it is related to the machine size?

chayim: Nope. Not 100% sure. Reduced as discussed back to medium. I think it's the remote docker issue (think, given then output).

DvirDukhan (approved):

PR #721, opened 2021-05-04
run cron twice aweek

gkorland:


DvirDukhan:
why twice a week?

rafie (dismissed):

.circleci/config.yml
```diff
@@ -564,16 +553,3 @@ workflows:
...
-  nightly:
-    triggers:
-      - schedule:
-          cron: "20 17 * * *"
-          filters:
-            branches:
-              only: master
-    jobs:
-#      - build-macos:
-#          <<: *never # temporarily disabled
```
rafie, line 577: Maybe keep a `dummy` + `<<: *never` just in case we want to do something nightly in the future? (it's impossible to have 0 jobs in a workflow, so dummy is needed if we want to keep the nightly clause).

DvirDukhan: Let's avoid "polluting" our files. If it is redundant, let's remove it.

rafie: @DvirDukhan Agree, but note that we're on a move towards an inter-module CircleCI form, so diverging from the common structure is not for the common good. But it's ok, I'll approve it.

rafie (approved):

PR #737, opened 2021-05-10
solving edge dockers overwriting each other

### LINKED ISSUE #736: Edge docker builds overwrite each other
RedisAI edge docker builds, overwrite each other - as they both push to edge-{cpu|gpu}. Instead, they should follow the versioning standard in place.

chayim:
merge after #730

closes #736


alonre24 (approved):
stampa

PR #738, opened 2021-05-10
beginning docker-compose migration

chayim:


DvirDukhan:
@chayim as we moved to `matrix` can we close this?

chayim:
It's in draft - I want to see if compose still makes sense for other reasons. Let me noodle on it.

PR #744, opened 2021-05-12
parallel platforms with remote dockers, matrix validation

chayim:
Changing CI to build all platforms, in parallel, for the platforms build. This also makes adding more in the future relatively easy, by editing the matrix variable.  We build a matrix of matrices - and run against those, in parallel.  We're also no longer using machines to build GPU docker images - but continue to for the **build_and_test_gpu** of course.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ commands:
             bash <(curl -fsSL https://get.docker.com)
             pushd opt/build/docker
             docker login -u redisfab -p $DOCKER_REDISFAB_PWD
-            make <<parameters.target>> OSNICK=<<parameters.platform>> <<parameters.lite>> ARTIFACTS=1 VERBOSE=1 build publish
+            make <<parameters.target>> OSNICK=<<parameters.osnick>> <<parameters.lite>> ARTIFACTS=1 VERBOSE=1 build publish
             popd > /dev/null
             logstar=bin/artifacts/tests-logs-cpu.tgz
             logsdir=tests/logs/cpu
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,7 +131,7 @@ commands:
     parameters:
       lite:
         type: string
-      platform:
+      osnick:
         type: string
       target:
         type: string
@@ -565,7 +565,7 @@ workflows:
           <<: *on-master-and-version-tags
           matrix:
             parameters:
-              platform:
+              osnic:
                 - xenial
                 - bionic
               lite:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,8 +130,8 @@ commands:
   platforms-build-matrix-steps:
     parameters:
       lite:
-        type: integer
-        default: 0
+        type: string
+        default: "LITE=1"
       platform:
         type: string
         default: "bionic"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -569,25 +547,33 @@ after-build-and-test: &after-build-and-test
 
 after-platform-builds: &after-platform-builds
```
rafie, line 548: I think we can replace `after-platform-builds` (and remove it altogether) with `platform-build`.

.circleci/config.yml
```diff
@@ -127,7 +127,14 @@ commands:
       - store_artifacts:
           path: tests/logs
 
-  platforms-build-steps-cpu:
+  platforms-build-matrix-steps:
+    parameters:
+      lite:
+        type: string
+      platform:
```
rafie, line 134: let's call it for what is it, `osnick`.

.circleci/config.yml
```diff
@@ -569,25 +547,33 @@ after-build-and-test: &after-build-and-test
...
-      - platforms-build-cpu:
-          <<: *after-build-and-test
-          <<: *on-master-and-version-tags
-      - platforms-build-gpu:
+      - platforms-build:
           <<: *after-build-and-test
           <<: *on-master-and-version-tags
+          matrix:
+            parameters:
+              platform:
```
rafie, line 568: let's call it for what is it, `osnick`.

.circleci/config.yml
```diff
@@ -233,22 +188,45 @@ jobs:
...
+        type: string
+      target:
+        type: string
     docker:
       - image: redisfab/rmbuilder:6.2.1-x64-buster
     steps:
-      - platforms-build-steps-cpu
+      - abort_for_docs
+      - abort_for_noci
+      - early_return_for_forked_pull_requests
```
rafie, line 204: Can we group these doggies together, in a command like `early-aborts` or something?

.circleci/config.yml
chayim: I prefer to keep it similar to the others. The other required elements are named **after-<foo>**. This matches those.

.circleci/config.yml
chayim: Pushed.

.circleci/config.yml
chayim: Pushed

.circleci/config.yml
chayim: Yes, but it's out of scope for this change. I'll batch it in something else. There's a branch open locally where that's in progress.

rafie (dismissed):

.circleci/config.yml
```diff
@@ -127,7 +127,14 @@ commands:
...
 
-  platforms-build-steps-cpu:
+  platforms-build-matrix-steps:
+    parameters:
+      lite:
+        type: string
+      osnick:
+        type: string
+      target:
+        type: string
```
DvirDukhan, line 137: add comments explaining what is what?

DvirDukhan: also, where is platform-build-matrix-steps is used?

chayim: Removed in push, but also documented variables in place with comments.

alonre24 (approved):

PR #747, opened 2021-05-13
restore readies master

chayim:
another branch (danglers) had been used while debugging disk size issues in cirlce.

given the parallel builds this is no longer needed


chayim:
@alonre24 This is a 'please stamp' situation. I'm removing the docker image removal that I introduced, since we no longer need it.

alonre24 (approved):

PR #753, opened 2021-05-20
adding build of pytorch on jetson - 1.7.0

chayim:
Fixed the build of pytorch so that it now works on Jetson hardware. I've pushed dependencies to s3.


alonre24 (approved):
I put stempa, but I also think that we should test these things in CI (i.e., running our test suit on jetson). Is it something that we have on our list?

PR #754, opened 2021-05-20
Aof rewrite fix and test

alonre24:
Refactoring and testing Redis AOF_Rewrite mechanism on ReidsAI types (probably for the first time...)

PR #764, opened 2021-05-30
removed aof tests

DvirDukhan:
as #754 introduced dedicated `aof` rewrite tests in the test files, the flag for `AOF` is now redundant and can be removed


rafie (dismissed):

.circleci/config.yml
```diff
@@ -249,7 +249,7 @@ jobs:
     parameters:
       test_args:
         type: string
-        default: "CLUSTER=0 AOF=0"
+        default: "CLUSTER=0"
```
chayim, line 252: If we're disabling the valgrind-aof job, we should change it's name in circleci to something more appropriate.

chayim (approved):

PR #766, opened 2021-05-31
circleci changes to meet the MOD-1227 goals with how ci runs

chayim:
1. valgrind-general will now run on external PRs, but not internal unless we add the **ci-test** label.
2. gpu builds will now run on external PRs, but not internal unless we add the **ci-test** label.
3. Runs on integ branches (master and versioned) continue to run gpu builds as before

This involves moving the executing steps from the workflows to the commands section, then reusing them. As such the workflow is duplicated (with rule changes and a new name), but the commands moved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,19 @@ commands:
               circleci step halt
             fi
 
+  only_run_if_forked_pull_request:
+    description: >-
+      If this build is from an internal (i.e not a forked) pull request, then end early.
+      This is required when we want different behaviours for external, versus internal PRs.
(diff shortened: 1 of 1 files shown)


chayim:
@DvirDukhan An interesting byproduct of this is that I (clearly) can't name the job, with the same name as during the non-PR run. This means that while we have GitHub rules requiring a passing gate for ci/circleci:build-and-test-gpu, in the forked context there's ci/circleci:build-and-test-gpu-for-forked-prs (etc).

DvirDukhan (approved):

PR #773, opened 2021-06-01
adding lite to the release automation

chayim:
Also added in to the nightly automation area of CircleCI, in case it gets used in the future


tests/qa/RS_VERSIONS-lite
```diff
@@ -0,0 +1,2 @@
+6.0.12-58
+100.0.0-2400
```
alonre24, line 2: This should be 100.0.0-2440 isn't it? @tomerhekredis ?

tests/qa/RS_VERSIONS
```diff
@@ -1,2 +1 @@
-6.0.12-58
 100.0.0-2400
```
alonre24, line 1: This should be 100.0.0-2440 isn't it? @tomerhekredis ?

tests/qa/run
```diff
@@ -89,10 +90,12 @@ run_test() {
 }
 
 rc=0
+# By default we use RS_VERSIONS, but with variants they may need to specify their own redis verisons, hence this file
+VERSIONFILE=${HERE}/RS_VERSIONS${VARIANT}
```
alonre24, line 94: We send `VARIANT=lite` from `circleci.yaml` but the file name is `RS_VERSIONS-lite` (with a dash) - is that a problem?

alonre24 (commented):
Few comments

tests/qa/run
chayim: CircleCI passes in VARIANT=-lite, which we use to build the file. This was validated with @tomerhekredis

tests/qa/RS_VERSIONS-lite
chayim: On non-lite @tomerhekredis said we wanted 100.0.0-2400 on on RS_VERSIONS-<variant> the 6.0x thing.  It's worth noting we can now basically do anything here by adding variants and files. @tomerhekredis ?

tests/qa/RS_VERSIONS
chayim: As above.

tests/qa/RS_VERSIONS-lite
tomerhekredis: yes, it should be 2440 in all RS_VERSIONS files

alonre24 (approved):

PR #778, opened 2021-06-06
Fix deadlock when running with gears

alonre24:
The RedisAI device run queue lock should not be locked when calling the DAG runInfo on_finish callback. The deadlock occurs whenever RedisAI background thread calls the asyncFinishCallback when running from gears via LLAPI while holding the queue lock, and it tries to acquire the python GIL. In parallel, RedisGears background thread is triggered and running a DAGRun call while holding the python GIL. The second thread is suspended waiting for the queue lock when it inserts the DAG runInfo to the queue - and that is how we get the deadlock.
The solution - do not hold the queue lock in RedisAI when running the on finish callback.


DvirDukhan (changes requested):
let's try to maintain a single lock-unlock scope
meaning:
if you get run info with error or time out, to a `goto` to the execution section, which unlocks the queue, then do a check if just to call `_BGThread_RinfoFinish` or execute and call `_BGThread_RinfoFinish`

after you re-obtain the lock, insert all unfinished run info

src/execution/background_workers.c
```diff
@@ -157,17 +157,27 @@ static bool _BGThread_IsRInfoTimedOut(RedisAI_RunInfo *rinfo) {
...
         }
     }
+
+    // Reinsert the unfinished DAG's run info to the queue.
+    pthread_mutex_lock(&run_queue_info->run_queue_mutex);
+    for (size_t i = 0; i < array_len(unfinished_rinfos_indices); i++) {
+        queuePushFront(run_queue_info->run_queue, batch_rinfo[unfinished_rinfos_indices[i]]);
+    }
+    pthread_mutex_unlock(&run_queue_info->run_queue_mutex);
+    array_free(unfinished_rinfos_indices);
```
DvirDukhan, line 180: Do this after the execution in the main loop. Keep a single lock-unlock scope.

DvirDukhan (changes requested):

DvirDukhan (changes requested):

DvirDukhan (approved):
very nice!

PR #781, opened 2021-06-08
tag fix

chayim:
Adding the docker tags on HEAD (i.e git tags), then fixing a force publish to dockerhub. This change is required in order to enable READIES to handle its tag + push flow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -575,7 +575,7 @@ workflows:
                 - xenial
                 - bionic
               lite:
-                - "REDISAI_LITE=0 publish"
+                - "REDISAI_LITE=0 publish PUSH_GENERAL=1 OFFICIAL=1"
                 - "REDISAI_LITE=1"
               target:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -575,7 +575,7 @@ workflows:
                 - xenial
                 - bionic
               lite:
-                - "REDISAI_LITE=0 publish"
+                - "REDISAI_LITE=0 publish PUSH_GENERAL=1 OFFICIAL=1"
```
DvirDukhan, line 578: why only in lite?
where is the branch/tag name instead of HEAD?

chayim: When you operate on a branch, you have a legitimate name. Tags are detached HEADs, hence this change. This change lands here specifically, because we only push to dockerhub when LITE=0, as before. This can change....

DvirDukhan (approved):

PR #783, opened 2021-06-09
new docker system

chayim:
RedisAI dockers, now generated from templates in the build tree. This is true of x64 cpu, gpu, and tests. This provides a way forward for templating in the future.  Makes use of the [new dockerwrapper](https://github.com/RedisLabsModules/readies/pull/42)


opt/build/docker/dockerfile-gpu-test.tmpl
```diff
@@ -13,25 +13,26 @@ ARG OS=ubuntu18.04
...
-
-RUN echo export LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda-11.0/lib64:$LD_LIBRARY_PATH > /etc/profile.d/cuda.sh
-
-COPY ./opt/ opt/
-COPY ./tests/flow/tests_setup/test_requirements.txt tests/flow/tests_setup/
-COPY ./tests/flow/tests_setup/Install_RedisGears.sh tests/flow/tests_setup/
+#COPY ./opt/ opt/
+#COPY ./tests/flow/tests_setup/test_requirements.txt tests/flow/tests_setup/
+#COPY ./tests/flow/tests_setup/Install_RedisGears.sh tests/flow/tests_setup/
+#COPY ./get_deps.sh .
```
DvirDukhan, line 35: remove

opt/build/docker/dockerfile-gpu-test.tmpl
```diff
@@ -13,25 +13,26 @@ ARG OS=ubuntu18.04
...
+FROM nvidia/cuda:{{REDIS_CUDA_VERSION}}-devel-ubuntu18.04 AS builder
+
+ARG PACK
 
 SHELL ["/bin/bash", "-c"]
 
 ENV NVIDIA_VISIBLE_DEVICES all
 ENV NVIDIA_DRIVER_CAPABILITIES compute,utility
+{% set cuda_list = REDIS_CUDA_VERSION.split("-") %}
+RUN echo export LD_LIBRARY_PATH=/usr/local/cuda/lib64:/usr/local/cuda-{{cuda_list[0]}}/lib64:$LD_LIBRARY_PATH > /etc/profile.d/cuda.sh
```
DvirDukhan, line 27: add comments expalining

chayim:
PR comments

- [ ] - documentation on building redisai in docker
- [ ] - comment on jinja sets and why
- [ ] - validate gpu test
- [ ] - split dockers into sub pieces now that they are reusable

DvirDukhan (commented):
add docs how to generate the docker files

rafie:
@chayim Note that Makefile at `/opt/build/docker` does not support `help`, so it isn't clear how to do something like `make build OSNICK=bionic` or anything else we used to support there.

chayim:
> @chayim Note that Makefile at `/opt/build/docker` does not support `help`, so it isn't clear how to do something like `make build OSNICK=bionic` or anything else we used to support there.

Given that these come from the only included file (docker.rules), I'd wager that the right place for the target there, as part of readies. They can always be overridden on a per-project basis, should that be necessary. After RTS (or something else), I'll migrate it there - once there's a pattern of 3.

The worrying issue is deploy-snapshots finding identical items. That, I'll look into first.

rafie:
@chayim I'm just reiterating a previous discussion: for the docker image generation process we have several args, among them: 
(1) osnick and architecture,
(2) either branch name or version (note that we cannot always deduce the branch name if it's not explicitly specified),
(3) is it an official build (that is, goes into redislabs/* repos),
(4) all kinds of content-related args, such as REDIS_VER, TEST, PACK and ARTIFACTS (not essential to the process - let's ignore those for now).
So given those args, we want a docker image generated and pushed into all relevant repos (redislabs/* and redisfab/* ones), possibly with different tagging for each repo - we'd like it to be a single, one-stop operation so users wouldn't need to think too much as they publish.
This should be supported inside a common infrastructure, as we don't want specific modules to contain logic but only configuration items like module name and names of integration branches.

In the current state of thing, we:
(1) support VERSION
(2) don't support BRANCH
(3) don't have standard docker repo policy (like `redisfab/mod:verspec-arch-osnick` and `redislabs/mod:verspec`).

chayim:
> (1) osnick and architecture,

This is set via the docker rules. They can be overiden and the help output has been modified accordingly for AI, but not for the architecture - until it's used.

> (2) either branch name or version (note that we cannot always deduce the branch name if it's not explicitly specified),

In AI we use versioned branches only, and tags. Setting the PUBLISH key on a per-project basis allows the projects to determine how and when they'd like to push, but that isn't the ask for this project.

> (3) is it an official build (that is, goes into redislabs/* repos)

Supported

> (4) all kinds of content-related args, such as REDIS_VER, TEST, PACK and ARTIFACTS (not essential to the process - let's ignore those for now).

Supported, though REDIS_VER is passed in to anything using the new system as REDIS_VERSION.

> So given those args, we want a docker image generated and pushed into all relevant repos (redislabs/* and redisfab/* ones), possibly with different tagging for each repo - we'd like it to be a single, one-stop operation so users wouldn't need to think too much as they publish.

Adding DOCKER_TAGS as specified to the help, on a per-project basis does this. By default the DOCKER_TAG is official - as that was the need here. If you'd prefer to push to redisfab/** by default, add the change to readies, and we can remove it from RedisAI.

> This should be supported inside a common infrastructure, as we don't want specific modules to contain logic but only configuration items like module name and names of integration branches.

This is already present. The makefiles only set variables - not targets.

> 
> In the current state of thing, we:
> (1) support VERSION
> (2) don't support BRANCH

Correct, but if you'd like to add support to it generically, that belongs in readies, not here. This project doesn't make use of branch builds that are non-versioned. The _docker.rules_ is the right place for that if necessary/

> (3) don't have standard docker repo policy (like `redisfab/mod:verspec-arch-osnick` and `redislabs/mod:verspec`).

We do, as defined in docker.rules. redisai has a separate naming convention given its variants, resulting in  the DEFAULT_TAG specific override for this project. The defaults are in the readies docker.rules.

rafie:
@chayim I'm looking on the overall experience, facing a "forklift upgrade" on all `/build/docker/Makefile`-s in all modules. 
Therefore, I look to match existing/planned interface from the command level, meaning we should have a set of commands that summarize the functionality of the system, like:
```
# latest version (this will usually take place on an integration branch):
make build OSNICK=bionic VERSION=1.2.3 LATEST=1

# latest version (on a branch off an integration branch):
make build OSNICK=bionic VERSION=1.2.3+fix1

# non-latest version
make build OSNICK=buster VERSION=1.2.2 PUBLIC=1
make build OSNICK=bionic VERSION=1.2.2

# public version
make build OSNICK=buster BRANCH=1.2 LATEST=1 PUBLIC=1
make build OSNICK=buster BRANCH=main EDGE=1 LATEST=1 PUBLIC=1 # should EDGE=1 be implicit here?

# branch builds (non-public)
make build OSNICK=bionic BRANCH=user-mypr1
make build OSNICK=bionic BRANCH=feature-f1
```
Let's try to extend the list until we run out of ideas. :-)

chayim:
> @chayim I'm looking on the overall experience, facing a "forklift upgrade" on all `/build/docker/Makefile`-s in all modules.
> Therefore, I look to match existing/planned interface from the command level, meaning we should have a set of commands that summarize the functionality of the system, like:
> 
> ```
> # latest version (this will usually take place on an integration branch):
> make build OSNICK=bionic VERSION=1.2.3 LATEST=1
> 
> # latest version (on a branch off an integration branch):
> make build OSNICK=bionic VERSION=1.2.3+fix1
> 
> # non-latest version
> make build OSNICK=buster VERSION=1.2.2 PUBLIC=1
> make build OSNICK=bionic VERSION=1.2.2
> 
> # public version
> make build OSNICK=buster BRANCH=1.2 LATEST=1 PUBLIC=1
> make build OSNICK=buster BRANCH=main EDGE=1 LATEST=1 PUBLIC=1 # should EDGE=1 be implicit here?
> 
> # branch builds (non-public)
> make build OSNICK=bionic BRANCH=user-mypr1
> make build OSNICK=bionic BRANCH=feature-f1
> ```
> 
> Let's try to extend the list until we run out of ideas. :-)

None of that belongs in RedisAI given how it's used. However currently, VERSION and OSNICK can be defined as part of the make call. I'd argue for placing this in readies, where it belongs - as opposed to the implementation of the tool. Given how RedisAI doesn't currently use the LATEST=/PUBLIC= I'd argue we should have that discussion in-context, elsewhere. It doesn't have this need as everything it pushes it public and there's no filtering. 

But, if you wanted to add an ifdef (PUBLIC) to the readies docker.rules and append to the docker tags, it would solve that. Similarly, a branch override would be welcome there - it's just that this product doesn't have that use case.

rafie:
@chayim All modules have the same requirements when it comes to docker build/deployment/naming and artifact creation/retrieval.

chayim:
> @chayim All modules have the same requirements when it comes to docker build/deployment/naming and artifact creation/retrieval.

Currently there's no external requirement placing the need for redisfab with RedisAI - though perhaps there is for other modules. What currently adds that requirement - the redisedge proposed change also pulls from the official redislabs/redisai docker, so these would align.  Naming wise, redisai similarly has changes with regards to platform variant (gpu/cpu) hence the name override in the Makefile.

I can be convinced of branch, but again this isn't a need we've encountered yet - as branches that product dockers are (currently) to be versioned. Either way, these would belong in readies and not RedisAI, given that this merely calls the dockerwrapper in that repository.

alonre24 (approved):

PR #785, opened 2021-06-15
onnxruntime 1.7.2 build and documentation

chayim:
Corrected how we build onnx. We can now build and release on x64, and x64 gpu. This will build on osx x64 as well, and version 1.8 of onnx now builds (using these changes) on jetson. This PR is only for 1.7.2. A new developer should be able to use the accompanying instructions, in order to work on the onnx build should they so desire.

This would be the template for building back-ends.  [ci was built in full](https://app.circleci.com/pipelines/github/RedisAI/RedisAI/3791/workflows/b6a8117e-985f-4b30-a627-1907a56a8a96) and the remaining run is for changing circleci back to master's norm.


opt/build/backends.rules
```diff
@@ -0,0 +1,48 @@
...
+		-t ${DEFAULT_DOCKER_TAG} \
+        -S ../dockerparts \
+		-e REDIS \
+		-D "${DOCKER_OPTS}" \
+		${DOCKER_ARGS}
+	docker create --cidfile ${CIDFILE} ${DEFAULT_DOCKER_TAG}
+	docker cp `cat ${CIDFILE}`:/build/${PRODUCT}/${BACKEND_NAME} .
+
+publish:
+	@aws s3 cp ${BACKEND_NAME} s3://$(S3_URL)/ --acl public-read
```
DvirDukhan, line 48: do we want to expose this?

opt/build/onnxruntime/Makefile
```diff
@@ -1,134 +1,16 @@
+REDIS_ONNX_VERSION?=rel-1.7.2_RCE
```
DvirDukhan, line 1: this is the branch name?

opt/build/onnxruntime/dockerfile.tmpl
```diff
@@ -0,0 +1,36 @@
+#----------------------------------------------------------------------------------------------
+{% if REDIS_ARCH == 'x64' %}
+{% if REDIS_GPU is defined %}
+FROM nvidia/cuda:11.0-cudnn8-devel-ubuntu18.04
```
DvirDukhan, line 4: maybe allow user to define their cuda version?
we use 11.0
might want to compile it on newer versions

alonre24:
moved to #847

opt/build/backends.rules
chayim: There's nothing being exposed here...

PR #791, opened 2021-06-17
Let ai.scriptget and ai.modelget commands run successfully without optional args

alonre24:
Align script get and model get command, so it will return both source/blob and meta if no optional argument is specified.
This PR also contains some documentation corrections for the new script commands.


DvirDukhan (commented):
please add comments to the `if` confitions or have some kind of format enumt

src/redisai.c
```diff
@@ -655,34 +655,31 @@ int RedisAI_ScriptGet_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv
...
-    }
-
+    // If only SOURCE arg was given, return only the script source.
     if (!meta && source) {
         RedisModule_ReplyWithCString(ctx, sto->scriptdef);
         return REDISMODULE_OK;
     }
+    // We return (META+SOURCE) if both args are given, or if none of them was given.
+    // The only case where we return only META data, is if META is given while SOURCE was not.
+    int out_entries = (source || !meta) ? 6 : 4;
```
DvirDukhan, line 675: I don't see what is the purpose of `META` here at all
it is either with or without source

DvirDukhan (dismissed):

DvirDukhan (approved):

PR #803, opened 2021-07-04
s/lite/light

chayim:


ramp-light.yml
```diff
@@ -1,10 +1,10 @@
-display_name: RedisAI-Lite
+display_name: RedisAILight
```
DvirDukhan, line 1: can it be `RedisAI Light`?

chayim: It is now. :)

DvirDukhan (approved):

PR #804, opened 2021-07-05
Update cluster tests

DvirDukhan:
this PR introduces better testing on OSS cluster
1. cluster tests are now running on 3 shards instead of 1
1. `AI.CONFIG` command is now "keyless" command
1. fixed a bug in the deprecated `SCRIPTRUN` command where the variadic symbol `$` was treated as key in `RedisAI_ScriptRun_IsKeysPositionRequest_ReportKeys`

as a result of this PR, all tests should use `get_connection(env, routing_hint)` to get a connection to the dedicated shard


alonre24 (commented):
Looks good! I had just one cosmetic comment....

alonre24 (approved):

PR #811, opened 2021-07-14
Update TensorSet and TensorGet in DAG upon execution 

alonre24:
Set status of dag tensorSet op at parse time (when it is executed). This should prevent race conditions that occur in DAG error whenever tensorset op is done in another device queue than the one where error occured.
Also, set tensorGet op status at reply time.

### DIFF
--- a/src/execution/DAG/dag.c
+++ b/src/execution/DAG/dag.c
@@ -480,31 +480,18 @@ void Dag_SetTensorInGlobalCtx(RedisAI_RunInfo *rinfo, size_t index, RAI_Tensor *
 void RedisAI_DagRunSessionStep(RedisAI_RunInfo *rinfo, const char *devicestr) {
     RAI_DagOp *currentOp = RedisAI_DagCurrentOp(rinfo);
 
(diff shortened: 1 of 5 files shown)


src/execution/DAG/dag.c
```diff
@@ -480,31 +480,18 @@ void Dag_SetTensorInGlobalCtx(RedisAI_RunInfo *rinfo, size_t index, RAI_Tensor *
...
-        break;
-    }
-    case REDISAI_DAG_CMD_TENSORGET: {
-        // TENSORSET op is done when we finish (consider removing it from dag ops).
-        currentOp->result = REDISMODULE_OK;
-        break;
-    }
-    case REDISAI_DAG_CMD_MODELRUN: {
+    // Verify that the op type belongs to the DAGCommand enum.
+    RedisModule_Assert(currentOp->commandType >= REDISAI_DAG_CMD_TENSORSET &&
```
alonre24, line 484: put a function that verifies the the range of the enum

DvirDukhan (approved):

PR #821, opened 2021-07-23
Investigate why TFLite auto batch test fail locally

alonre24:
Try to see what CI is getting and compare it to my local results


alonre24:
Problem solved when upgrading numpy locally to 1.19

PR #877, opened 2021-11-21
Add a reference to RedisGears-RedisAI integration from AI docs

alonre24:


alonre24:
Use #878 instead (same PR but with _docs suffix, so it won't run CI)

PR #884, opened 2021-11-30
Fix device name validation in "modelstore" command

alonre24:
The fifth character was not validated.
Also, for scriptstore and the deprecated modelset and scriptset, the only validation occurs in `parseDeviceStr` (which is from within the backends, after parsing stage). This PR replaces `sscanf` with `strtoll` in `parseDeviceStr`, to have safe exit in case of invalid device id string in these commands.

### DIFF
--- a/src/redisai.c
+++ b/src/redisai.c
@@ -173,12 +173,14 @@ int RedisAI_ModelStore_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **arg
     const char *devicestr;
     AC_GetString(&ac, &devicestr, NULL, 0);
     bool valid_device = false;
+    size_t device_prefix_len = strlen("GPU:"); // can also be "CPU:"
     if (strcasecmp(devicestr, "CPU") == 0 || strcasecmp(devicestr, "GPU") == 0) {
         valid_device = true;
-    } else if ((strncasecmp(devicestr, "GPU:", 4) == 0 || strncasecmp(devicestr, "CPU:", 4) == 0) &&
-               strlen(devicestr) <= 10) {
+    } else if ((strncasecmp(devicestr, "GPU:", device_prefix_len) == 0 ||
+                strncasecmp(devicestr, "CPU:", device_prefix_len) == 0) &&
+               strlen(devicestr) < 10) {
         bool digits_only = true;
-        for (size_t i = 5; i < strlen(devicestr); i++) {
+        for (size_t i = device_prefix_len; i < strlen(devicestr); i++) {
             if (devicestr[i] < '0' || devicestr[i] > '9') {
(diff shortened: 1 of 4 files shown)


src/redisai.c
```diff
@@ -173,12 +173,14 @@ int RedisAI_ModelStore_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **arg
...
     AC_GetString(&ac, &devicestr, NULL, 0);
     bool valid_device = false;
+    size_t device_prefix_len = strlen("GPU:"); // can also be "CPU:"
     if (strcasecmp(devicestr, "CPU") == 0 || strcasecmp(devicestr, "GPU") == 0) {
         valid_device = true;
-    } else if ((strncasecmp(devicestr, "GPU:", 4) == 0 || strncasecmp(devicestr, "CPU:", 4) == 0) &&
-               strlen(devicestr) <= 10) {
+    } else if ((strncasecmp(devicestr, "GPU:", device_prefix_len) == 0 ||
+                strncasecmp(devicestr, "CPU:", device_prefix_len) == 0) &&
+               strlen(devicestr) < 10) {
```
DvirDukhan, line 181: why the change from `<=` to `<`?

tests/flow/tests_commands.py
```diff
@@ -19,13 +19,22 @@ def test_modelstore_errors(env):
...
+
+    # Check for valid device argument (should only be "CPU:<n>" or "GPU:<n>, where <n> is a number
+    # (contains digits only, up to 5).
     check_error_message(env, con, "Invalid DEVICE",
                         'AI.MODELSTORE', 'm{1}', 'TORCH', 'bad_device', 'BLOB', model_pb)
+    check_error_message(env, con, "Invalid DEVICE",
+                        'AI.MODELSTORE', 'm{1}', 'TORCH', 'CPU::', 'BLOB', model_pb)
+    check_error_message(env, con, "Invalid DEVICE",
+                        'AI.MODELSTORE', 'm{1}', 'TORCH', 'CPU:1.8', 'BLOB', model_pb)
+    check_error_message(env, con, "Invalid DEVICE",
```
DvirDukhan, line 36: add a comment explaining why this is invalid device

DvirDukhan (commented):
where was the 5th character did not validated?
anyway, looks good

src/redisai.c
alonre24: No actual reason... I believe that 5 digits should be enough

tests/flow/tests_commands.py
alonre24: There is a comment in lines 28-29 explaining it...

DvirDukhan (approved):

PR #888, opened 2021-12-18
updated pytorch and cuda version

DvirDukhan:
this PR updated to pytroch 1.10.1 (1.10.1-cuda11.3 on GPU)

### DIFF
--- a/opt/system-setup.py
+++ b/opt/system-setup.py
@@ -30,7 +30,7 @@ class RedisAISetup(paella.Setup):
         self.install("gawk")
         self.install("libssl-dev")
         self.install("python3-regex")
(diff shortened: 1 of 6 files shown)


opt/system-setup.py
```diff
@@ -30,7 +30,7 @@ def debian_compat(self):
         self.install("gawk")
```
chayim, line 30: I know this sounds silly - but let's just add all of the self.install to a single call:
```
self.install('libssl-dev python3-regex python3-networkx python3-numpy')
```

The reason, is that it calls apt once, meaning this will run faster.

chayim (approved):

PR #892, opened 2021-12-26
avoid magic numbers in tflite

DvirDukhan:
this PR fixed inconsitancy in tflite test. 
it validates both outputs are the same and not nan or inf


alonre24 (approved):
Nice

PR #894, opened 2021-12-30
added 120 seconds timeout for env restart and reload

DvirDukhan:


GuyAv46 (approved):
Looks good
Should we dig into why it takes so long to load?

PR #919, opened 2022-05-29
Use debug symbols

alonre24:
Add `-g` and `-ggdb` flags to the compiler, so we have the ability to get the debug symbols using gdb from the module's binary in production.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,9 +22,9 @@ ENDIF()
 
 #----------------------------------------------------------------------------------------------
 
-SET(CMAKE_CC_COMMON_FLAGS "-fPIC -fcommon")
+SET(CMAKE_CC_COMMON_FLAGS "-fPIC -fcommon -g -ggdb")
 IF (USE_PROFILE)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -O3")
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3")
 
 # Add -fno-omit-frame-pointer to avoid seeing incomplete stack traces
-set(CMAKE_COMMON_FLAGS_DEBUG "-g -ggdb -fno-omit-frame-pointer -D_DEBUG -DVALGRIND -include \
+set(CMAKE_COMMON_FLAGS_DEBUG "-fno-omit-frame-pointer -D_DEBUG -DVALGRIND -include \
```
DvirDukhan, line 45: don't you want to chain the existing `CMAKE_CC_COMMON_FLAGS` with this?

PR #924, opened 2022-06-02
Set default backends path

alonre24:
Set the backends' default path on module load, so it can be retrieved with AI.CONFIG GET command. Also - pass the module's path to flow tests.


src/config/config.c
```diff
@@ -150,7 +149,7 @@ int Config_SetIntraOperationParallelism(RedisModuleString *num_threads_string) {
 int Config_SetModelChunkSize(RedisModuleString *chunk_size_string) {
     long long val;
     int result = RedisModule_StringToLongLong(chunk_size_string, &val);
-    if (result != REDISMODULE_OK || val <= 0) {
+    if (result != REDISMODULE_OK || val <= 0 || val > REDISAI_DEFAULT_MODEL_CHUNK_SIZE) {
```
DvirDukhan, line 152: why the addition of this condition?

src/config/config.h
```diff
@@ -28,6 +28,8 @@ typedef enum { RAI_DEVICE_CPU = 0, RAI_DEVICE_GPU = 1 } RAI_Device;
 #define REDISAI_INFOMSG_MODEL_EXECUTION_TIMEOUT "Setting MODEL_EXECUTION_TIMEOUT parameter to"
 #define REDISAI_INFOMSG_BACKEND_MEMORY_LIMIT    "Setting BACKEND_MEMORY_LIMIT parameter to"
 
+#define REDISAI_DEFAULT_MODEL_CHUNK_SIZE (511 * 1024 * 1024)
```
DvirDukhan, line 31: why the addition?

src/redisai.c
```diff
@@ -1424,6 +1429,8 @@ int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 
     RedisModule_SetModuleOptions(ctx, REDISMODULE_OPTIONS_HANDLE_IO_ERRORS);
 
+    RedisModule_SubscribeToServerEvent(ctx, RedisModuleEvent_Shutdown, RAI_CleanupModule);
```
DvirDukhan, line 1432: what is the minimal redis version supporting this event? does it align with our min_redis_version on our ramp files?

tests/flow/tests_onnx.py
```diff
@@ -567,7 +567,7 @@ def test_synchronization(self):
 
         def launch_redis_and_run_onnx(con, proc_id, pipes):
             my_pipe = pipes[proc_id]
-            port = 6380 + proc_id  # Let every subprocess run on a fresh port.
+            port = 6380 + 30*proc_id  # Let every subprocess run on a fresh port (safe distance for RLTEST parallelism).
```
DvirDukhan, line 570: didn't you remove RLTest parallelism?

src/redisai.c
alonre24: yes, this API is supported from version 6.0.0, which is the minimal version for RedisAI anyway....

src/config/config.c
alonre24: We need to limit the `chunk_size`, since Redis cannot reply with blobs that are larger than 512MB, right?

src/config/config.h
alonre24: To use this constant in `Config_SetModelChunkSize`

src/backends/backends.c
```diff
@@ -92,33 +92,17 @@ int RAI_ExportFunc(const char *func_name, void **targetFuncPtr) {
...
-
-    return backends_path;
+    // Retrieve the info about the module's dynamic library, and extract the .so file dir name.
+    RedisModule_Assert(dladdr(RAI_SetBackendsDefaultPath, &info) != 0);
+    const char *dyn_lib_dir_name = dirname((char *)info.dli_fname);
+
+    // Populate backends_path global string with the default path.
+    size_t backends_default_path_len = strlen(dyn_lib_dir_name) + strlen("/backends");
+    *backends_path = RedisModule_Alloc(backends_default_path_len + 1);
+    RedisModule_Assert(sprintf(*backends_path, "%s/backends", dyn_lib_dir_name) > 0);
```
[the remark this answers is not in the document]
alonre24: Is that better? We don't use `HoldString` or something here (so I can't see why using `RedisModuleString` object has an advantage...)

tests/flow/tests_onnx.py
alonre24: Yes... But I don't think that this addition is bad, as we might wan't to use the parallelism in the future. WDYT?

src/config/config.c
DvirDukhan: https://github.com/redis/redis-doc/pull/1653#issuecomment-922891303
see here. You can configure the server to support bigger values

DvirDukhan (approved):

PR #925, opened 2022-06-06
Fetch test data from s3 instead of using git-lfs

alonre24:
Remove git-lfs from this project, and use S3 to store tests data (which includes some heavy models). We use `s3://dev.cto.redis` bucket, and restore data from it in CI using sync command. The data is also stored in circle-CI local cache


.circleci/config.yml
```diff
@@ -198,6 +218,23 @@ commands:
       - restore_cache:
           keys:
           - v1.2.5-deps-{{ checksum "get_deps.sh" }}-gpu
+      - restore_cache:
+          keys:
+            - v1.2-tests_data-gpu
```
DvirDukhan, line 223: what is the reason we are having an additional cache key? Should be the same for both modes, no?

tests/flow/tests_setup/tests.sh
```diff
@@ -94,6 +86,17 @@ valgrind_summary() {
 
 #----------------------------------------------------------------------------------------------
 
+get_tests_data() {
```
DvirDukhan, line 89: I don't mind having this right now, but it forces developers to have aws account to run this command. 
Please see https://stackoverflow.com/questions/273743/using-wget-to-recursively-fetch-a-directory-with-arbitrary-files-in-it if we can use `wget`  with aws HTTP.

.circleci/config.yml
alonre24: You're right, that's how it was before, but I got `access denied` when restoring cache, so I tried this option instead....

tests/flow/tests_setup/tests.sh
alonre24: Done. Using recursive `wget` was not possible, so we got the desired behavior by using a file that specify all the relevant resources to download.

DvirDukhan (approved):

PR #936, opened 2022-09-09
Update ORT_ARCH for Mac OS X

ashao:
As of v1.10.0 of the ONNX Runtime, Microsoft changed the string associated with the processor architecture in the name of the Mac OS X release tarball. The variable ORT_ARCH now correctly specifies x86_64 instead of x64, allowing Mac OS X users to retrieve the necessary dependency.


DvirDukhan (approved):

DvirDukhan:
@ashao thanks!

PR #940, opened 2022-11-27
Removing LGTM

chayim:
The LGTM service is being shut off in two weeks. This pull request aims to remove all references to LGTM. Perhaps LGTM usage should be replaced with codeql, or a repository preferred tool, but IMHO that's the point of a different pull request.


ankit-gautam23 (approved):
LGTM

